PR #47, opened 2018-10-17
PAR - Kevin Rabesaotra pas fini

knandraina:


starter-code/src/contact.js
```diff
@@ -0,0 +1,67 @@
...
+            contact: contacts.sort((a, b) => a.name > b.name)//Ne pas mettre de slice à cet endroit, sinon on bloque l'array à une certaine taille
+        })
+    }
+
+    sortByPopularity = () => {
+        this.setState({
+            contact: contacts.sort((a, b) => a.popularity > b.popularity)
+        })
+    }
+
```
lucamerzi: The logic is there but I have the impression that your SortByPopularity() is not working. Try doing something like this when sorting numbers:
```js
function compareNumbers(a, b) {
  return a - b;
}
```

starter-code/src/contact.js
```diff
@@ -0,0 +1,67 @@
...
+        this.setState({
+            clickCount: this.state.clickCount + 1
+        })
+    }
+    sortByName = () => {
+        this.setState({
+            contact: contacts.sort((a, b) => a.name > b.name)//Ne pas mettre de slice à cet endroit, sinon on bloque l'array à une certaine taille
+        })
+    }
+
```
lucamerzi: When sorting alphabetically you have to return a negative number to keep the same order and a positive one to invert it (see example below). Also, don't forget to uppercase or lowercase, in case two elements start by the same letter and sort needs to look at the second item:
```js
// sort by name
items.sort(function(a, b) {
  var nameA = a.name.toUpperCase(); // ignore upper and lowercase
  var nameB = b.name.toUpperCase(); // ignore upper and lowercase
  if (nameA < nameB) {
    return -1;
  }
  if (nameA > nameB) {
    return 1;
  }

  // names must be equal
  return 0;
});
```

starter-code/src/contact.js
```diff
@@ -0,0 +1,67 @@
...
+    sortByPopularity = () => {
+        this.setState({
+            contact: contacts.sort((a, b) => a.popularity > b.popularity)
+        })
+    }
+
+    deleteSelectedElement = (movieIndex) => {
```
lucamerzi: good logic 😎👍 !

ta-webpt-par:
Well done overall, just make sure you revise your sorting methods 😉 : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

PR #167, opened 2019-02-20
Paris:Poeli

pbojork:
I think I forgot all of module 1 :/


starter-code/src/components/Ironcontacts.js
```diff
@@ -0,0 +1,62 @@
...
+    // update the array
+    const contact = this.state.contactsArray;
+    contact.push(newContact);
+
+    // setState() to tell React to changen the DOM);
+    this.setState({ contactsArray: contact });
+  }
+
+  randomContact() {
+    const randomIndex = Math.floor(Math.random() * contactsArray.length);
```
ta-web-paris: Here you're using again the contactsArray from your constructor so you have to write it with `this.state` like this : 
```
const randomIndex = Math.floor(Math.random() * this.state.contactsArray.length);
this.addContact(this.state.contactsArray[randomIndex]);
```

ta-web-paris:
Continue to work hard Poeli :)
If you can I encourage you to go to the reinforcement of tonight (which is on this lab) :)
It'll cover the stuffs we need on the module 1 too, so you'll find it really useful.
Otherwise the correction will be available at the end of the week and the code of Abi's reinforcement will be available tomorrow, try to spend time on it and try to finish it by yourself :)

PR #180, opened 2019-02-24
[WEBPT-PARIS] Done. 

mernri:
@ta-web-paris 
Hello 👋
I display the 5 first artists in a table, but it looks like the picture, the name and the popularity of each artist are in the same column. 

**EDIT :**  The problem is fixed. the reason behind it was that i was enclosing the < tr > tag in a < div >. So i just had to remove the div.


ta-webpt-par:
Great job !

PR #202, opened 2019-03-04
FTWD MIA Cam Chromiak/ Matt Keil

cameron-chromiak:
finished


ta-web-mia:
Good job! The only small thing that I feel should be fixed is that your sorting algorithm for popularity is sorting the lowest popularity to the top, rather than the highest. To remedy this, simply switch the lines of code on lines 38 and 41 of the App.js like so:

image

Overall your app satisfies every requirement, and you've done a good job implementing all of the functions. Keep up the good work 👍

PR #204, opened 2019-03-04
FTWD Miami Jan 2019

MikeTroianello:
iterations 1 and 2 finished


ta-web-mia:
So far so good! Although the directions recommend that you implement a sorting algorithm both alphabetically and by popularity, the delete button is actually a lot more straightforward. All it requires you to do is to splice a celebrity at a given index. You're on the right track with all of this, keep it up 👍

PR #205, opened 2019-03-04
Victoria Hickam & Nunna FTWD MIAMI

vhickam:
Done


ta-web-mia:
Great job! Just a minor issue, but the instructions wanted you to sort the highest popularity to the top instead of the lowest to the top. All you need to do to fix it is to switch the if statements within the sortByPopularity function like so:

image

No real problems with any of the methods though, everything works great! Keep up the good work 👍

PR #206, opened 2019-03-04
Manuel Garaboa / Monroe Getz FTWD-Miami

mannygaraboa:
First Commit


ta-web-mia:
Well done! The only thing that can be improved upon, other than styling, is that the "Sort by name" button is sorting them alphabetically by first name instead of last name. Overall though, very good job. Keep up the good work 👍

PR #208, opened 2019-03-07
Henry Doce - FTWD - Miami

hendo12:


ta-web-mia:
Good job! All of the buttons are working just as intended. Hopefully you feel a little more caught up now that you have a better understanding of props and state. Those are two of the biggest things to grasp, so stay focused and keep up the good work 👍

PR #229, opened 2019-03-25
iterations 1, 2, and 4 done.

williamfiallos:


williamfiallos:
done during class code along.

ta-web-mia:
Well done! I would just suggest adding a check to your add random contact button. Check to make sure you're not adding any duplicates

PR #230, opened 2019-03-27
[PT1018MIA]Camilo Campo

mcpccampo:


ta-web-mia:
Well done! A suggestion for later is to have the load random celeb check to make sure it doesn't load duplicates

PR #237, opened 2019-04-30
PAR alex

alexandfox:


starter-code/src/components/Contact.js
```diff
@@ -0,0 +1,103 @@
...
+		console.log("rendering.... state contacts: ", this.state.contacts)
+		// if (this.state.sort === "name") this.filteredContacts = this.sortByName()
+		// else if (this.state.sort === "popularity") this.filteredContacts = this.sortyByPopularity()
+		// else this.filteredContacts = this.state.contacts;
+
+		return (
+			<div className="contactContainer">
+				<header className="title">IronContacts</header>
+				<button className="button randomButton" onClick={() => this.addRandomContact()} >Add Random Contact</button>
+				<button className="button" onClick={() => this.sort("name")} >Sort by Name</button>
```
ta-web-paris: it's cool that you're using the same function twice, and while it works in this example. It won't always work. The sort method works based on numbers, for sorting alphabetically it's better to localeCompare because uppercase and lowercase letters have different values -  https://www.w3schools.com/charsets/ref_html_ascii.asp

ta-web-paris:
I really love your code <3 Everything was done really well

PR #243, opened 2019-04-30
Paris Yagmur 

YagmurBakici:


starter-code/src/App.js
```diff
@@ -1,21 +1,80 @@
...
+  sortName() {
+    this.state.allContacts.sort(function(celeb1, celeb2) {
+      if (celeb1.name < celeb2.name) return -1;
+      if (celeb1.name > celeb2.name) return 1;
+      else return 0;
+    });
+    this.setState({ listContacts: this.state.allContacts });
+  }
+
+  delete = name => {
```
ta-web-paris: your delete doesn't work because you've called the function here delete, but the function that is linked to your button is `deleteInfo` and your state should be `    this.state.allContacts.splice(name, 1);`

ta-web-paris:
It looks good, and I really like the styling! You were super close with getting the delete button to work and you seem to have understood passing down props 🤚

PR #244, opened 2019-04-30
Paris, Ciao Ragazzi

MiguelIronHack:


starter-code/src/components/Contacts.js
```diff
@@ -0,0 +1,97 @@
+import React, { Component } from "react";
+import contacts from "../contacts.json";
+
+export class Contacts extends Component {
+  state = {
+    firstFive: contacts.slice(0, 5)
+  };
+
+  addNew = () => {
+    const random = Math.floor(Math.random() * 200);
```
ta-web-paris: multiplying by 200 works this time, but the length of the array could change so it's better to times it by the length of the `contacts`

starter-code/src/components/Contacts.js
```diff
@@ -0,0 +1,97 @@
...
+    const randomOne = contacts.slice(random, random + 1);
+    const currContact = [...this.state.firstFive, ...randomOne];
+    this.setState({ firstFive: currContact });
+  };
+  sortByName = () => {
+    const sortMap = [...this.state.firstFive];
+    const sortedMap = sortMap.sort((a, b) => (a.name < b.name ? -1 : 1));
```
ta-web-paris: for sorting alphabetically it's better to use `localeCompare`  because the sort method is based using numbers and lowercase and uppercase letters have different values

starter-code/src/components/Contacts.js
```diff
@@ -0,0 +1,97 @@
...
+  };
+  sortByPop = () => {
+    const sortMap = [...this.state.firstFive];
+    const sortedMap = sortMap.sort((a, b) =>
+      a.popularity > b.popularity ? -1 : 1
+    );
+    this.setState({ firstFive: sortedMap });
+  };
+  deleteCurr = e => {
+    console.log("The UK is not part of the real Europe.");
```
ta-web-paris: I am very disappointed by this statement.

ta-web-paris:
It looks good and everything works as expected 👍

PR #274, opened 2019-05-11
[AMS PT] - Fedde

Fedduh:
Done


starter-code/src/ContactOverview.js
```diff
@@ -0,0 +1,76 @@
...
+
+        <p className="button-action" onClick={this.createRandomContact}>Add random contact</p>
+        <p className="button-action" onClick={this.sortContactByName}>Sort by name</p>
+        <p className="button-action" onClick={this.sortContactByPopularity}>Sort by popularity</p>
+
+        {/* Contact table (bit hacky with delete button here.. but otherwise need uplift) */}
+        <ContactTable>
+          {this.state.currentContacts.map(contact => {
+            return (
+              <ContactRow key={contact.name} name={contact.name} pictureUrl={contact.pictureUrl} popularity={contact.popularity}>
```
TA-Ironhack-Matt: It's better to use the {index} as a value for the key, because with the index you are 100% sure that the value will be unique. If there would be a duplicate name in in the data this can cause errors. 

the `.map()` function by default returns an index. You can acces this index by writing:
`this.state.currentContacts.map((contact, index) => { ......`

TA-Ironhack-Matt: The other props `name={contact.name} pictureUrl={contact.pictureUrl} popularity={contact.popularity}` that you are passing to `<ContactRow/>`  are all properties of the element. 

This can be done with fewer code when using the spread operator like this
 `<ContactRow key={contact.name} {...contact}>`

starter-code/src/ContactOverview.js
```diff
@@ -0,0 +1,76 @@
...
+    let newContacts = this.state.currentContacts.splice(selectedContact, 1)
+    this.setState({ currentContact: newContacts })
+  }
+
+  render() {
+    return (
+      <div>
+        <h1>{this.props.title}</h1>
+
+        <p className="button-action" onClick={this.createRandomContact}>Add random contact</p>
```
TA-Ironhack-Matt: A small tip would be, to move all buttons to a separate component called for example `<TableControls/>`. when you want to extend you a table with new controls you can add all that logic to the `<TableControls/>` component and use your `<ContactOverview/>` component as a component that constructs the overview from separate components.

PR #278, opened 2019-05-12
done

jelenajukic:


starter-code/src/Description.js
```diff
@@ -0,0 +1,18 @@
+import React, { Component } from 'react';
+import "./Description.css"
+
+class Description extends Component {
```
TA-Ironhack-Matt: Try to be a bit more descriptive when naming your components to make your code better readable for other developers. "Description" is not a very clear name for this component, for example "ContactRow" would be more clear in this case. :)

PR #280, opened 2019-05-13
[AMS] Merle Jansen

jurlem:


starter-code/src/contactsList.js
```diff
@@ -1,4 +1,4 @@
-[
+const contacts = [
```
TA-Ironhack-Matt: Following the assignment, this data was supposed to be stored in a .json file and not in a .js file. 

But, I like the your approach on seperating the logic for selecting the first five apart from your components.js. This approach can come in very handy when we will be working with lifecycle methods :)

starter-code/src/contactsList.js
```diff
@@ -994,4 +994,13 @@
     "pictureUrl": "https://image.tmdb.org/t/p/w500/aqZ9RjL5j44HMlBMvTaawhHiGOH.jpg",
     "popularity": 6.914606
   }
-]
\ No newline at end of file
+]
+
+
+export function getContacts() {
```
TA-Ironhack-Matt: As mentioned before, very nice to export functions like this and call them in any component where you need it. 👍

starter-code/src/Contacts.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+  handlePopular = () => {
+    let contacts = this.state.contacts.sort((a,b) => {
+     return a.popularity > b.popularity ? -1 : a.popularity < b.popularity ? 1 : 0
+  })
+    this.setState({contacts})
+  }
+
+
+  render () {
```
TA-Ironhack-Matt: This component is to big. Try to split it up in smaller components, for example in the following components: 
- TableControls />
- ContactRow />

starter-code/src/Contacts.jsx
```diff
@@ -0,0 +1,74 @@
...
+    return (
+      <div>
+        <h1>IronContacts</h1>
+        <button onClick={this.handleAddRandom} type="button" className="btn btn-dark ">Add random Contact</button>
```
TA-Ironhack-Matt: These buttons could be encapsulated in the TableControls component that I mentioned before :)

starter-code/src/Contacts.jsx
```diff
@@ -0,0 +1,74 @@
...
+            <tr>
+              <th>Picture</th>
+              <th>Name</th>
+              <th>Popularity</th>
+              <th>Action</th>
+            </tr>
+          </thead>
+          <tbody>
+            {this.state.contacts.map ((contact, index) => (
+              <tr key={index}>
```
TA-Ironhack-Matt: Would be nice to just call a component for example "ContactRow" (mentioned before) and pas in the data to it each iteration of the Map function. 

This would make your code more readable.

starter-code/src/Contacts.jsx
```diff
@@ -0,0 +1,74 @@
...
+import {getContacts} from './contactsList';
+
+class Contacts extends Component {
+  state = {
+    contacts: getFirstFive (),
+    fulllist: getContacts()
+  };
+
+  handleDelete = contact => {
+    //let contacts = [...this.state.contacts];
```
TA-Ironhack-Matt: Overall your methods are very clean and well written, well done! 

Small remark on this line is to remember to remove all out-commented code before you hand in the assignment.

PR #283, opened 2019-05-14
MIA FTWB April Parker Cooper

parkcoop:


ta-web-mia:
Nice job! I did experience one minor issue though, where duplicate celebrities were being added to the list. You could probably add a conditional within the function to check if the celebrity already exists in the current list, and then call the function inside of itself to make it recursive. Just an idea, and nothing major enough to get hung up on. Keep up the good work!

PR #287, opened 2019-05-14
MIA FTWD APR - Katy Brown

katybrown91:


ta-web-mia:
Nice job! Minor issue with duplicates showing up when contacts are added, but nothing to get too hung up over. Keep up the great work 👍

PR #290, opened 2019-05-14
MIA FTWD APR Daniela Olano

dolano13:


ta-web-mia:
Nice job! I'd suggest putting the buttons on the top of the page rather than the bottom. No real issues though, keep up the great work 👍

PR #304, opened 2019-05-23
[MAD PT] Guille LA

GuilleLA:
Good exercise to start getting use to react.

I made a lot of mistakes organising the structure of the app i had to build as  a consecuence i got to a point in wich is kind of complicated to access the data, plus there are some bugs with the keys. 

I now know how should i have done it. Therefore, i will continue with the next exercise and will only get back to this one if i have time.


ta-web-mad:
Muy bien Guille, no ser vago y pon las keys !! 😄 

#checked #closed #Giorgio 🇮🇹

PR #339, opened 2019-07-16
Berlin-Adrienne Gispen-Stefan Culafic-FTWD

adrigispen:


starter-code/src/components/ContactsList.js
```diff
@@ -0,0 +1,75 @@
...
+    contactsCopy.splice(contactIndex, 1);
+    this.setState({
+      contacts: contactsCopy
+    });
+  };
+
+  addContact = () => {
+    let rand = Math.floor(Math.random() * this.props.contacts.length);
+    let contact = this.props.contacts[rand];
+    const contactsNew = this.state.contacts;
```
mjarraya: This is a copy by reference. Prefer copying the values
```suggestion
    const contactsNew = [...this.state.contacts]; // or .slice()
   
```

starter-code/src/components/ContactsList.js
```diff
@@ -0,0 +1,75 @@
...
+    contactsNew.unshift(contact);
+    this.setState({ contacts: contactsNew });
+  };
+
+  sortByName = () => {
+    let sortedByName = this.state.contacts.sort((a, b) => {
```
mjarraya: .sort() mutates the array in place. Protect your array against mutations ;)

starter-code/src/components/ContactsList.js
```diff
@@ -0,0 +1,75 @@
+import React, { Component } from "react";
+import Contact from "./Contact";
+
+class ContactsList extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      contacts: props.contacts.splice(0, 5),
+      sort: "none"
```
mjarraya: This doesn't seem to be used.

starter-code/src/components/ContactsList.js
adrigispen: 👍

starter-code/src/components/ContactsList.js
adrigispen: yeah, I thought I might need it for the deletion but then realized that it was rerendering the whole thing so each elem always already had its proper index

PR #367, opened 2019-07-23
MIA-FTWD-Sergio

sergiovelo17:


ta-web-mia:
Nice work, and I really like that you added a delete random button. It's always great to go and implement extra features, even if they aren't required, especially if it helps you learn something new! Keep up the great work 👍 
- Nick

PR #368, opened 2019-07-23
MIA-FTWD-Artem/William

arstrel:
Artem and William: friends with React


ta-web-mia:
- Nick

PR #456, opened 2019-10-01
[MAD DE]Marta Rodriguez

Marta-RoDi:


ta-web-mad:
“You might belong in Hufflepuff,
Where they are just and loyal,
Those patient Hufflepuffs are true,
And unafraid of toil.”
― J. K. Rowling, Harry Potter and the Sorcerer's Stone

"Muy bien!!!"
-Carlos Trujillo

PR #459, opened 2019-10-01
MIA-FTWD-Lauren L.

justlask:
started out too complex with too many components, but also learned about how to pass up functionality from 3-4 levels deep so thats a win. not going for bonus styling.


ta-web-mia:
Lookin' solid. Good work on the state management.

PR #463, opened 2019-10-01
mia ftwd aug

VitorTomazzi:


ta-web-mia:
Everything looks good! I'm glad you're getting the hang of React, it's a lot to take in at once, but it makes sense the more you use it, and you got everything working on this.

PR #464, opened 2019-10-01
MIA-FTWD-AUG2019-Jimmy Thomas

hoshimotao:


ta-web-mia:
Really nice job on this! I know you've been struggling with React, but this is nigh on perfect, for one of the hardest labs in the whole bootcamp. Keep going, you're almost there!

PR #521, opened 2019-11-13
done :)

jnguyen3436:


ta-web-mia:
Great job on the lab. If you get the chance try and get the array to update so you don't have copies of the same contact. Keep up the good work! ~ Stefan

PR #525, opened 2019-11-14
FTWD MIA - October 2019 - Pradeepa Shanmugasundaram

pradeepa24:


ta-web-mia:
Great job on the lab.
If you get the chance try and change the array so you don't get multiple copies of same contact. Keep up the great work! ~ Stefan

PR #533, opened 2019-11-26
[PAR-WD]Camille

Millcat:
All done except bonus


ta-web-paris:
Good job - you reversed the order for sort by popularity but the logic is here so all good :)

PR #536, opened 2019-11-26
[PAR-FTWD]-LOUISE

LouiseMarissal:
done but some troubles with the delete button need explainations plz
Thank you !


ta-web-paris:
Almost perfect, but just make sure you sort by name/popularity only on the contacts you're already displaying, not on all of them (it adds all the contacts from the database when you sort), you just need to use the function on the correct array, but no big deal :))

PR #620, opened 2020-01-17
[AMS FT 1119] Hannah Hawes

hanny242:


starter-code/src/components/Celebrities.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import './Celebrities.scss'
+
+const Celebrities= (props) => {
```
elyasp: Shorten the code by doing:

``` export default const Celebrities = ( ..... etc ```

PR #656, opened 2020-02-18
[WDFT-PAR0120] Andrew && Cécile ongoing

kamacess:


kamacess:
Andrew +

TatijanaRajcic:
Hey guys! Thanks for your lab! It looks liiiiiiiit ! 

Some remarks:
- you could "merge" your two sort function:

```
    handleSort = (criteria) => {
        const copy = [...this.state.displayedContacts];
        var sortedCopy = copy.sort((a, b) => (a[criteria] > b[criteria]) ? 1 : -1)
        this.setState({ displayedContacts: sortedCopy })
    }
```

```
 <button id= "name" onClick={()=>this.handleSort("name")}>Sort by name</button>
 <button id= "popularity" onClick={()=>this.handleSort("popularity")}>Sort by popularity</button>
```

- it's nice that you tried to add some logic to the randomAdd to avoid adding a contact that is already in the first 5. But then with this logic, you are not sure that you'll need get twice the same person in your reservedContacts array. Maybe you can set the state of reservedContacts each time (by removing the randomContact) ? That's just an idea :)

- you could simplify your handleDelete function. As you associate an event handler to your onclick event listener, you could pass, to your function, the index itself. Therefore, you will not need the event anymore;

`<td><button className="delete-btn" id = {i} onClick = {()=>this.handleDelete(i)}>X</button></td>
`
```
    handleDelete = (index) => {
        var copy = [...this.state.displayedContacts];
        var UpdatedCopy = copy.filter((celeb, i) => index !== i)
        this.setState({ displayedContacts: UpdatedCopy})
    }
```

Good job anyway :))

PR #667, opened 2020-02-19
 [PAR WDFT]- Shannon, iterations 1-3 done ongoing

tranquildoe:


TatijanaRajcic:
Hey Shannon!

Thanks for your lab:) it's really good :) It's cool that you made the effort to create a component MyContact, instead of having everything in the app.js

Small remarks to make it even better;
- you can merge your two sort methods in one:

```
    filter = (criteria) => {
        const copy = [...this.state.celebsToDisplay];
        copy.sort((a, b) => (a[criteria] > b[criteria]) ? 1 : -1)
        this.setState({ celebsToDisplay: copy })
    }

```
```
 <button onClick={()=>this.filter("name")}>Sort by name</button>
 <button onClick={()=>this.filter("popularity")}>Sort by popularity</button>
```

- try to add the delete button and come and see me if you are not sure how to do it :). you'll need the pass a props to your component myContact, that triggers when the button from this component is clicked, and when this function is called, it is going to call a method from the parent component, which is going to change de celebsToDisplay . Give it a try and come and see me if you have any question :)

good job!

PR #668, opened 2020-02-19
[BER-FT-022020] Jacqueline C. & Karo C.

chenjacqueline:
Done!

### DIFF
--- a/starter-code/src/Contacts.js
+++ b/starter-code/src/Contacts.js
@@ -1,0 +1,140 @@
+import React, { Component } from "react";
+import contacts from "./contacts.json";
+
+class Contacts extends Component {
+  state = {
+    firstFive: [
+      {
+        name: "Idris Elba",
+        pictureUrl:
+          "https://image.tmdb.org/t/p/w500/d9NkfCwczP0TjgrjpF94jF67SK8.jpg",
+        popularity: 11.622713,
+        id: "11731993-0604-4bee-80d5-67ad845d0a38"
+      },
+      {
+        name: "Jessica Chastain",
+        pictureUrl:
+          "https://image.tmdb.org/t/p/w500/nkFrkn5NZVGWb4b2X0yIcXezhyt.jpg",
+        popularity: 8.324357,
+        id: "17980511-75ca-48b0-bea8-462fec2ee43d"
+      },
+      {
+        name: "Johnny Depp",
+        pictureUrl:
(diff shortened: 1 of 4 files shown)


starter-code/src/Contacts.js
```diff
@@ -0,0 +1,140 @@
+import React, { Component } from "react";
+import contacts from "./contacts.json";
+
+class Contacts extends Component {
+  state = {
+    firstFive: [
```
fenrew: This is a bit hard-coding right here, it would be cleaner to just slice the first 5 contacts from your imported contacts

starter-code/src/Contacts.js
```diff
@@ -0,0 +1,140 @@
...
+        pictureUrl:
+          "https://image.tmdb.org/t/p/w500/A85WIRIKVsD2DeUSc8wQ4fOKc4e.jpg",
+        popularity: 11.245333,
+        id: "b4d2c7b8-fdd5-426a-85bd-011c3f50a6c6"
+      }
+    ]
+  };
+
+  randomContact = () => {
+    let index = Math.floor(Math.random() * contacts.length);
```
fenrew: Math.floor(Math.random() * contacts.length-5)+5
This will prevent you from re-picking your first 5 contacts

starter-code/src/Contacts.js
```diff
@@ -0,0 +1,140 @@
...
+    let newContact = contacts[index];
+
+    this.state.firstFive.forEach(contact => {
+      if (contact.id === newContact.id) {
+        let index = Math.floor(Math.random() * contacts.length);
+        newContact = contacts[index];
+      }
+      return newContact;
```
fenrew: I don't think this return does anything? Since it's a .forEach.

starter-code/src/Contacts.js
```diff
@@ -0,0 +1,140 @@
...
+      }
+    ]
+  };
+
+  randomContact = () => {
+    let index = Math.floor(Math.random() * contacts.length);
+    let newContact = contacts[index];
+
+    this.state.firstFive.forEach(contact => {
+      if (contact.id === newContact.id) {
```
fenrew: Do I understand it correctly what you are trying to do here is that if you already have the contact in your contacts-list, then get a new random contact? 

If yes: Keep in mind that if the contact is already in your this.state.firstFive, then you change it, but only once! What you would have to do is to put this into a while loop that doesn't stop until you get a contact that is not found in your this.state.firstFive

PR #670, opened 2020-02-19
WD-FT Berlin Vaso and Alexis 

Perdikal:


starter-code/src/App.js
```diff
@@ -1,18 +1,113 @@
...
+const firstFive = [];
+for (let i = 0; i < 5; i++) {
+  firstFive.push(contacts[i]);
+}
+//console.log(contacts.length);
+
+//console.log(randomNumber);
+class Table extends Component {
+  state = {
+    firstContacts: firstFive
```
fenrew: Better to use contacts.slice(0,5)

starter-code/src/App.js
```diff
@@ -1,18 +1,113 @@
...
+  };
+  add = () => {
+    let randomNumber = Math.floor(Math.random() * 199 + 5);
```
fenrew: instead of 199, use contacts.length to make it dynamic, just keep in mind that you have to dubtract 5 afterwards to not go out of the array:

Math.floor(Math.random() * contacts.length + 5)-5

starter-code/src/App.js
```diff
@@ -1,18 +1,113 @@
...
+    this.state.firstContacts.push(contacts[randomNumber]);
+    this.setState({
+      firstContacts: this.state.firstContacts
+    });
+  };
+
+  sortByName = () => {
+    this.setState({
+      firstContacts: this.state.firstContacts.sort((a, b) => {
+        if (a.name.toLowerCase() < b.name.toLowerCase()) {
```
fenrew: If they are equally bit it is supposed to return 0. localeCompare is the best practice for comparing strings though

starter-code/src/App.js
fenrew: Also keep in mind that you never check if the contact is already in your this.state.firstContacts meaning you may get an error with your key={elem.id} as the id may thus be similar for some contacts

PR #671, opened 2020-02-19
[BER -FT] Laurel & Sha

lwrocks:
Will keep working through iterations 2 and 3.

### DIFF
--- a/starter-code/src/App.js
+++ b/starter-code/src/App.js
@@ -40,6 +40,24 @@ class App extends Component {
     });
   };
 
+  deleteContact = index => {
+    // when we receive an index
+    console.log(this.state.contactsLists);
+    console.log(JSON.parse(JSON.stringify(this.state.contactsLists)));
+    console.log("delete contact", index);
+    // let currentContact = JSON.parse(JSON.stringify(this.state.contactsLists));
+    // let newContact = currentContact.splice(index, 1);
+
+    this.state.contactsLists.splice(index, 1);
+    // splice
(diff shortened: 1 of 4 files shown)


starter-code/src/App.js
```diff
@@ -1,18 +1,74 @@
...
+  addRandomContact = () => {
+    // here we find a random contact
+    // and we add it to the state for contactsList
+
+    const random = contacts[Math.floor(Math.random() * contacts.length)];
+
+    console.log(random);
+
+    this.setState({
+      contactsLists: [random, ...this.state.contactsLists]
```
fenrew: Keep in mind that you never check if the contact is duplicated meaning you may get an error when you set your Contact Component key={id} since this id may be similar for some of the contacts

starter-code/src/App.js
```diff
@@ -1,18 +1,74 @@
...
+    });
+  };
+
+  sortByName = () => {
+    this.setState({
+      contactsLists: this.state.contactsLists.sort((a, b) => {
+        if (a.name > b.name) return 1;
```
fenrew: Nice! Don't forget you also have LocaleCompare

starter-code/src/App.js
```diff
@@ -1,18 +1,74 @@
...
+        else return -1;
+      })
+    });
+  };
+
+  sortByPopularity = () => {
+    this.setState({
+      contactsLists: this.state.contactsLists.sort((a, b) => {
```
fenrew: You dont have to specify what to return here. This is sufficient: .sort((a,b) => a.popularity - b.popularity)

PR #672, opened 2020-02-19
Berlin FT "almost done" Kiyo & Lluis

LluisManas:


starter-code/src/contacts.js
```diff
@@ -0,0 +1,91 @@
...
+    });
+    console.log(this.randomContact());
+    console.log(this.state.contacts, "contacts");
+  };
+
+  toggleName = () => {
+    let sortedNames = this.state.contacts;
+    console.log(this.state.contacts);
+    sortedNames.sort((a, b) => (a.name > b.name ? 1 : -1));
+
```
fonsogms: maybe here you could have return the reversed sorted array?
Right now is only sorting in ascending order.

PR #684, opened 2020-02-22
[PAR-WDFT] Kathleen. [ongoing]

kathd:


starter-code/src/App.js
```diff
@@ -26,19 +26,54 @@ export default class App extends Component {
...
+
+  sortByPopularity = () => {
+    this.setState({
+      displayedContacts: [...this.state.displayedContacts].sort((a,b) => {
+        if (a.popularity > b.popularity) return -1;
+        if (a.popularity < b.popularity) return 1;
+        return 0;
+      })
+    })
+  }
```
frankmarmier: The issue here with having your filters defining the state is that when you add a new contact, the filters are not applied unless you click again on the filters.

You are free to implement some filters before the JSX template or  set a function that returns a filtered array so it gets called on every rerender, hence taking care of newly added contacts.

frankmarmier:
Well done !

PR #723, opened 2020-02-26
[BCN-FTWD01/20-VALENTIN]

valentin-muller:
Only finished the first few iterations because I couldn't continue on my own.


valentin-muller:
Finished the last iterations with Henriette.

PR #826, opened 2020-04-29
[AMSWDPT202001] Daniel Hoogerwerf

danielhoogerwerf:


starter-code/src/App.js
```diff
@@ -1,18 +1,87 @@
...
+            <thead>
+              <tr>
+                <th>Picture</th>
+                <th>Name</th>
+                <th>Popularity</th>
+              </tr>
+            </thead>
+            <tbody>
+              {this.state.contacts.map((element, key) => (
+                <tr key={key}>
```
lotofcaffeine, line 64: You could improve this part creating a functional component just to encapsulate the information from the <tr>

starter-code/src/App.js
```diff
@@ -1,18 +1,87 @@
...
+          <table>
+            <thead>
+              <tr>
+                <th>Picture</th>
+                <th>Name</th>
+                <th>Popularity</th>
+              </tr>
+            </thead>
+            <tbody>
+              {this.state.contacts.map((element, key) => (
```
lotofcaffeine, line 63: It is not good to use the index in a list. Reacts works better when the key is unique, you could use element.id instead.

PR #834, opened 2020-05-01
done

fhoman:


starter-code/src/App.js
```diff
@@ -1,18 +1,88 @@
...
 
 class App extends Component {
+
+  state = {
+    contacts: contacts.splice(0, 5),
+  };
+
+
+  deleteContact = this.deleteContact.bind(this);
+
```
LloydC, line 17: An alternative way to use your deleteContact without using "bind" would be `<RowTable key={index} deleteContact={() => this.deleteContact(index)} />` then in your RowTable.jsx you would need to update your code like this `<button onClick={props.deleteContact}`

PR #836, opened 2020-05-05
[PAR-WDTF-0320] – Rebecca Norén ☎️ongn

rebeccanoren:
I'm done! 💯


ta-web-paris:
Hi Rebecca !
Great job !

One small issue is with the delete : it doesn't delete the one I want, but another one, and I checked your code and it looks good to me, i don't understand (I'll ask Tatijana and Franck if they spot something)

in the mean time I know this method works :
>> removeContact = (index) => {
    this.setState({contacts: this.state.contacts.filter((u, i) => i !== index)});
  };

I'll send you a message regarding what we found !!
Nina

TatijanaRajcic:
Hey! You have an issue while deleting because:

- You are mapping over the sorted copy of your contacts. Let's say I open your app, I do "sort by name", the first celebrity I have is Emily Clarke. Her index, inside what you return from sortContact, is 0), so if I click on her delete button, the handleDelete will take 0 as an argument.
- But then, in the delete method, you use this index to retrieve an element, from the ORIGINAL array (i mean, a copy of it, but it's still refer to the un-sorted contact list). Therefore, you are deleting the element with 0 from the original array, which is Idris Elba.
- You have several ways to solve this and actually using filter with the id here would not solve your issue either. But you could use the filter with the name! The names are the same, being in the sorted or unsorted lists.

```
<button onClick={(event) => this.handleDelete(contact.name)}>
  Delete
</button>
```

```
  handleDelete = (contactName) => {
    this.setState({
      contacts: this.state.contacts.filter(
        (contact) => contact.name !== contactName
      ),
    });
  };
```

This is the way to make it work with the way you constructed it. 

Otherwise, you would have to rethink about your logic, and keep track of the sorted list in your state as well (because right now you don't keep track of your sorted list anywhere since it only depends on a "display" logic).

Hope it helps!

PR #883, opened 2020-05-09
PAR WDFT March 2020 Sébastien DUBOIS

sdubois90:


ta-web-paris:
Good job Sebastien !
One small thing : i would have sorted the popularity from most to less popular (makes more sense)

Otherwise everything works well !

Nina

PR #1024, opened 2020-08-18
[WDFT-REMOTE-0820] Alejandro

alejandrohj:


IH-WebDev-TA-Remote:
Very good solution, keep it up!

PR #1025, opened 2020-08-18
[WDFT-REM-2007] Britta Helm

brittahelm:


IH-WebDev-TA-Remote:
Code looks good. Main feedback would be to use the reduced version of the sort method. Also you can do a direct splice on the contacts to get the 5 elements instead of doing some code inside the constructor. Kudos on using another component for the contact info :)

PR #1026, opened 2020-08-18
[WDFT-REM-2007] Elise Dujardin

elisedjn:
All done! I tried to do only one function for the both sort button and it seems to work :)


IH-WebDev-TA-Remote:
Oh nice! very clever solution with the dynamic sort method :) and even some styles, very good job!

PR #1027, opened 2020-08-18
[WEB-REM-2007] Lanette Lanting

llanting:
All done! I included some minimal styles just to try it. Also, I ended up changing my list into a table because I liked to look of that better.


IH-WebDev-TA-Remote:
Good solution and even some styles. Good job! one suggestion would be to use reduced versions of the sort methods.

PR #1028, opened 2020-08-18
[DEV-REM-2007] Bodei 

BodeiBrouwer:


IH-WebDev-TA-Remote:
Code looks good! great work! Only suggestion would be to use reduced syntax for the sort methods.

PR #1029, opened 2020-08-18
[WDFT-REM-2007] Maeva Merrou

maevamerrou:
Lab finished


IH-WebDev-TA-Remote:
Good job! one suggestion would be to use reduced syntax for the sort methods. Keep it up! :)

PR #1030, opened 2020-08-18
[WDFT-REM-2007] --- LAB-react-ironcontacts ---    Joel Roca Martinez

JoelRocaMartinez:


IH-WebDev-TA-Remote:
Good job! one suggestion would be to get used to the reduced syntax versions of the sort methods.

PR #1031, opened 2020-08-18
[WEB-REM-2007] ALVARO

AlvaroSLam:
Lab finished


IH-WebDev-TA-Remote:
Good job! one suggestion would be to get used to the redux syntax of the sort methods.
Also, nice work creating a new component for the contact information!

PR #1032, opened 2020-08-18
React lab tuesday

GunnerAg:
I wanted to separate all the components in files, but I started with the bonus lab !


IH-WebDev-TA-Remote:
Very good job! I think working with only that extra component for contact info is enough here, you will have more situations when dividing into more components is needed. But again, very cool solution and even some styles. Keep it up!

PR #1033, opened 2020-08-18
[WEB-REM-2007] Marta Lourido

MartaLourido:
done without the bonus

At first I was a bit stuck with the first interaction, but once I get it, the last ones came faster and seemed similar to what we did in the morning.

I tried to add styles to the buttons but it didn't work, maybe because I have to import the CSS in Table.js, but as when I did it I got an error I decided to leave it like that.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,13592 @@
+{
+  "name": "starter-code",
+  "version": "0.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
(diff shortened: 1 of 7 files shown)


IH-WebDev-TA-Remote:
Very good job! having that extra component for the contact info was a good touch. Don't worry about styles as that was not the important part of the lab here. One suggestion would be to use reduced syntax for the sort methods and get confortable with that as it is used often in projects. But again, code looks good and easy to follow. Kudos!

PR #1034, opened 2020-08-18
[WEB-REM-2007] Rafael

rlaz133:
I would have loved giving it some styles, but I could not finish the extra lab on time.


IH-WebDev-TA-Remote:
Very cool solution! lots of components to I see you are already confortable using them and even passing props. One suggestion would be to use reduced syntax for the sort methods.
But again, very cool job, even without styles :P

PR #1038, opened 2020-09-10
[AMS - WEBDEV - 072020] - Alex Anthopoulos

AlexAnthopoulos:


reMRKableDev (commented):
Nice work @AlexAnthopoulos 🚀 
There is a ```console.log``` on line 31 in Contacts.js that can be removed

PR #1127, opened 2020-10-06
[PAR-WDFT-248] MAXIMILIAN GRUBER

MaxJGruber:


N10T:
Hello Max !

Good job and amazing CSS 💯 

My proposal
-

sort are modifying the array itself, you don't need to save it in another constant
```
  sortByName = () => {
    const copy = [...this.state.contacts];
    const sortedCopy = copy.sort((a, b) => {
      if (a['name'] > b['name']) {
        return 1;
      } else if (a['name'] < b['name']) {
        return -1;
      } else {
        return 0;
      }
    });

    this.setState({
      contacts: sortedCopy,
    });
  };
```

Can be : 
```
  sortByName = () => {
    const copy = [...this.state.contacts];
    copy.sort((a, b) => {
      if (a['name'] > b['name']) {
        return 1;
      } else if (a['name'] < b['name']) {
        return -1;
      } else {
        return 0;
      }
    });

    this.setState({
      contacts: copy,
    });
  };
```

PR #1128, opened 2020-10-06
[PAR - WDFT - 248] Timothée NICOLE

timothee-nicole:
iteration 1 to 4 done, not the bonus yet
Have a good correction !

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,13808 @@
+{
+  "name": "starter-code",
+  "version": "0.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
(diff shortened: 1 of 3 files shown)


N10T:
Yo Tim'

Good job 💯 don't worry for the bonus.

My proposal
-

filter is to "filter" an array not to create another one, prefer forEach in this case or reduce, if you just need a loop
```
const fiveContact = []
Contacts.filter((elem, i) => {
  if (i <= 4) {
    fiveContact.push(elem)
  }
})
```

Can be : 
```
const fiveContact = Contacts.filter((elem, i) => i < 4)
```

Or easier:
```
const fiveContact = Contacts.slice(0,5)
```

PR #1129, opened 2020-10-06
[PAR-WDFT-248]Bey Bouziane

Bouziane95:


N10T:
Yo Bouziane,

Good job and nice CSS 💯  but ...

In your console you can see that: 
`Warning: Each child in a list should have a unique "key" prop.`
and
`<tr> cannot appear as a child of <table>`

My proposals
-

For the first:
```
<Contacts
              key={contact.id} //I add a unique key
              id = {contact.id}
              name={contact.name}
              popularity={contact.popularity.toFixed(2)}
              picture={contact.pictureUrl}
              deleteContact={this.deleteContact}
            />
```

For the second: https://developer.mozilla.org/fr/docs/Web/HTML/Element/table 👍

PR #1132, opened 2020-10-06
[PAR-WDFT-248] Jorell Langmar

JorellLangmar:


N10T:
Hello Jorell,

Good job and nice popping CSS!

`Warning: Each child in a list should have a unique "key" prop.` 
This is warning is because you forget the key :
```
LINE 71 :{this.state.AllContact.map((person, i) => {
                return (
                  <tr key= {i}>
```

PR #1133, opened 2020-10-06
[PAR-WDFT-248] Laura Caillaux

Lcdevi:


N10T:
Yo Laura,


Good job!

You have 2 warnings due to this:

You wrote this line 92 : 
```
<tr>
key={i}
```
React want a parent tag with a prop "key" with a unique value
`<tr key={i}>`

PR #1136, opened 2020-10-06
[WDFT PAR 248] Vincent Proum

VProum:


N10T:
Yo Vincent,

Good job !

You don't need to use "new Set()" evrytim when you don't need to take off the duplicates

PR #1151, opened 2020-10-27
[REM-2009] Sumiya Ushiro

sumi0820:
Completed all iteration. Please review it.


IH-WebDev-TA-Remote:
Amazing work with using material styles and working around hooks. But ultimately there are several issues with the button functionalities. 

1. In the hooks, starting with only 5 contacts, not all:

`const [newContacts, setNewContacts] = useState(contacts.splice(0,5));`

2. The add button only needed to add a new contact to the current list. So a functionality like this:

````
const handleRandom = () => {
    const randomContacts = [...newContacts];
      const randomIndex = Math.floor(Math.random() * randomContacts.length);
      randomContacts.push(contacts[randomIndex]);
    setNewContacts(randomContacts);
  };
````

3. Finally, the sort methods should work with the current state (newContacts) of the app and sort those, not bringing all of the original data (contacts):

`let sortByName = [...newContacts];`

Check instructions closely for the next Lab since it will be a little similar but with added touches. And even if right now working with hooks is complicated, you can very well use class components and continue learning about hooks later :) Knowing both is very important in the long run.

PR #1155, opened 2020-10-27
[REM 2009] Ross Homaidan

ibirossi:
I tried to do the table and some styling but it went very wrong.  So, I reverted back to the basic lab.


IH-WebDev-TA-Remote:
Good job, zero styles but the functionalities (the most important bit) are all there and working correctly! Keep it up!

PR #1156, opened 2020-10-27
[REM 2009] Vykintas Petrikas

VykiPetr:


IH-WebDev-TA-Remote:
Functionalities are there but check how the setState method is used. You should never directly modify the state of the component via push or sort or any other way. Always creating a new element and use the setState method to update the state with that element.

for example: instead of:

````
handleSortPop = () => {
        this.state.table.sort((a, b) => {
            if (a.popularity < b.popularity) {
                return 1
            } else if (a.popularity > b.popularity) {
                return -1
            } else {
                return 0
            }
        })
        this.setState({})
    }
````

do:

````

handleSortPop = () => {
        let newArr = [...this.state.table]
        newArr.sort((a, b) => {
            if (a.popularity < b.popularity) {
                return 1
            } else if (a.popularity > b.popularity) {
                return -1
            } else {
                return 0
            }
        })
        this.setState({table:newArr})
    }
````


Apply this to all the functionalities.
Other thank that, all good logic and working correctly.

PR #1157, opened 2020-10-27
[ REM 2009 ] Jordane Ruiz

jordaneruiz:
all iterations done. did everything alone after iteration 2 :))))))


IH-WebDev-TA-Remote:
Awesome! congratz! :) it looks good and all the functionalities working correctly. The only feedback would be to create more components so you get used to working like that. It will be essential later on when working on more complex labs and your project. But still, very well done! keep it up!

PR #1158, opened 2020-10-27
[REM WDFT 2009]Marta Obregon

MartaObregon:


IH-WebDev-TA-Remote:
Well done! all functionalities working correctly. Keep it up!

One small feedback, when writing your if/else conditionals in single lines, you don't need the { }

````
if(a.name>b.name) return 1
else if(a.name<b.name) return -1
else return 0
````

Similar for methods, if you are writing them in a single line you don't need the {} or even the return keyword.

PR #1159, opened 2020-10-27
[ 2009 REM ] Aleksandra Tanasijenko

alekstana:
Yoohoooo


IH-WebDev-TA-Remote:
Good job! all functionalities working correctly, I would only suggest you get used to dividing your code into more components, it will help you a lot on the long run for more complex labs and specially your projects. Still, very good job, keep it up!

PR #1200, opened 2020-11-25
[FT-AMS-102020] Elia Chiarucci

eliachiarucci:


LloydC:
Good work Elia,

There's nothing to add, your solution works and is clear.

PR #1201, opened 2020-11-25
[FTAMS102020] Nina van den Braak

ninsajala:


src/App.js
```diff
@@ -1,24 +1,92 @@
...
+
+function ShowContacts() {
+  const [contacts, setContacts] = useState(Contacts.splice(0, 5));
+
+  const randomContactHandler = () => {
+    const contactCopy = [...contacts];
+    const getRandomContact = () =>
+      Contacts[
+        Math.floor(
+          Math.random() * (Contacts.length - (contactCopy.length - 1)) +
```
wax-bear, line 13: Something is off in this logic.
Imagine: there are 10 Contacts, and 15 contacts in contactCopy. A Math.random would be multiplied with a negative number. Right?

src/App.js
```diff
@@ -1,24 +1,92 @@
...
+            (contactCopy.length - 1)
+        )
+      ];
+    contactCopy.push(getRandomContact());
```
wax-bear, line 17: Okay, this looks good.
On line 6, the Contacts are spliced that are not supposed to be in the array anymore.
Could that also be applicable here? Or is it desirable to have the same actor multiple times in the list?

src/App.js
wax-bear: Please check the next comment. Taking that comment into account, you can retrieve the actor randomly quite a bit easier.

src/App.js
```diff
@@ -1,24 +1,92 @@
-import React from 'react';
-import logo from './logo.svg';
+import React, { useState } from 'react';
 import './App.css';
+import Contacts from './contacts';
+
+function ShowContacts() {
+  const [contacts, setContacts] = useState(Contacts.splice(0, 5));
+
+  const randomContactHandler = () => {
```
wax-bear, line 8: I would name this (even) more descriptively: "addRandomContactHandler".

src/App.js
```diff
@@ -1,24 +1,92 @@
...
+  const sortPopularityHandler = () => {
+    let contactCopy = [...contacts];
+
+    contactCopy = contactCopy.sort(function (a, b) {
+      return b.popularity - a.popularity;
+    });
+    setContacts(contactCopy);
+  };
+
+  const deleteHandler = (contactId) => {
```
wax-bear, line 47: Great stuff!
I have one question: if you delete a contact, should it be possible to add it randomly later?
If so, the deleted contact should be added to Contacts.

wax-bear (commented):
Great stuff Nina!
Please check out the comments for potential improvements.
If you have a question, please reach out.

src/App.js
ninsajala: Thanks for the feedback Thomas! There were indeed some things off, I think I resolved them now (or at least made it better)

PR #1205, opened 2020-11-25
[FT-AMS-102020] Emad Ghnim

Emad-doz:


src/Components/DynamicTable.js
```diff
@@ -0,0 +1,77 @@
...
+                <td>{popularity.toFixed(2)}</td>
+                <td><button onClick={()=>this.delete(index)}>Delete</button></td>
+            </tr>
+        )
+        });
+    }
+
+    render() {
+        return (
+        <div class='style'>
```
LloydC, line 54: You should be using `className` instead of `class` here.

src/Components/DynamicTable.js
```diff
@@ -0,0 +1,77 @@
...
+        <h1 id='title'>IronContacts</h1>
+        <button onClick ={() => this.addRandom()}>Add Random Contacts</button>
+        <button onClick={()=>this.sortByName()}>Sort By Name</button>
+        <button onClick={()=>this.sortByPopularity()}>Sort By Popularity</button>
+        <table class = "tableStyle" id='ID'>
```
LloydC, line 59: Same comment as above.

src/Components/DynamicTable.js
```diff
@@ -0,0 +1,77 @@
...
+class DynamicTable extends Component{
+
+    state = {contactsState: contacts.slice(0,5)}
+
+    addRandom(){
+        let randomContact = contacts[Math.floor(Math.random()*contacts.length)]
+        this.setState({contactsState: [...this.state.contactsState, randomContact]})
+    }
+
+    sortByName(){
```
LloydC, line 14: Hey Emad, this works fine but for the sake of making your code more readable I would recommend writing your function in a way where instead of having  the sorting happen at the same time as you update the state breaking this in two separate actions
```
sortByName(){
       let sortedContacts = this.state.contactsState.sort((a, b) => { 
            let nameA = a.name.toUpperCase(); 
            let nameB = b.name.toUpperCase(); 
            if (nameA < nameB) {
              return -1;
            }
            if (nameA > nameB) {
              return 1;
            }
            return 0;
        })

        this.setState({contactsState: sortedContacts})     
    }

File: src/Components/DynamicTable.js [hunk 4]
```diff
@@ -0,0 +1,77 @@
...
+            if (nameA < nameB) {
+              return -1;
+            }
+            if (nameA > nameB) {
+              return 1;
+            }
+            return 0;
+        })})     
+    }
+
```
LloydC, line 27: Same rhetoric here
```
sortByPopularity(){
        let sortedContacts = this.state.contactsState.sort((a, b) => b.popularity - a.popularity);
        this.setState({contactsState: sortedContacts});
    }

PR #1209, opened 2020-11-26
[FTAMS102020] Emma Northridge

emmanorthridge:
Done but haven't added CSS

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,38 +1,100 @@
+import React, { Component } from 'react';
 import './App.css';
 import contacts from './contacts.json';
-import React, { useState } from 'react';
 
-function App() {
-  const [contactList, setContacts] = useState(contacts.slice(0, 5));
-
-  const randomContact = () => {
-    const copyContact = [...contactList];
-    const random = contactList[Math.floor(Math.random() * contactList.length)];
-    copyContact.push(random);
-    setContacts(copyContact);
+class App extends Component { 
+  state = {
+    contactList: contacts.slice(0, 5),
   };
 
-  return (
-    <div className="contactsList">
-      <button onClick={randomContact}>Add another contact</button>
-      <table>
-        <tr>
-          <th>Picture</th>
-          <th>Name</th>
-          <th>Popularity</th>
-        </tr>
-        {contactList.map((item) => (
-          <tr key={item.id}>
-            <td>
-              <img src={item.pictureUrl} alt="contact" />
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -1,26 +1,100 @@
...
-  );
+class App extends Component { 
+  state = {
+    contactList: contacts.slice(0, 5),
+  };
+
+  addRandom = () => {
+    const copyList = [...this.state.contactList];
+    const random = contacts[Math.floor(Math.random() * contacts.length)];
+    if (!copyList.includes(random)) {
```
wax-bear, line 13: Please be mindful that this `.includes` works because copyList and random are both derived from contacts, so the objects in these variables have the same object references.

To exemplify that:

1. [{a: 1}].includes({a:1}); // false
2. var arr = [{b:2}]; var arrCopy = arr; arrCopy.includes(arr[0]); // true

https://stackoverflow.com/a/6605700/3563737

wax-bear: The logic can be improved. 
If a user clicks the "Add another contact" button, we would like a contact to be added in any case, right?
If a match is found, perhaps another random contact can be retrieved, compared to the copyList and added if there is no match.
This repetitive operation could be stored in a function. For example "getRandomContact".

wax-bear (commented):
Great stuff Emma!
Please check my comments regarding the addRandom method and the table HTML tags.
Perhaps some CSS could improve the styling of the page.

src/App.js
```diff
@@ -1,26 +1,100 @@
...
+  render() {
+    return (
+      <div className="App">
+        <button onClick={() => this.addRandom()}>Add another contact</button>
+        <button onClick={() => this.sortByName()}>Sort by name</button>
+        <button onClick={() => this.sortByPopularity()}>
+          Sort by popularity
+        </button>
+        <table>
+          <tr>
```
wax-bear, line 73: This table row can be nested in a `thead`.

src/App.js
```diff
@@ -1,26 +1,100 @@
...
+            <th>Picture</th>
+            <th>Name</th>
+            <th>Popularity</th>
+          </tr>
+          {this.state.contactList.map((mapContact) => {
```
wax-bear, line 78: The following table rows can be nested in a `tbody`.

wax-bear: Also, "Each child in a list should have a unique "key"", according to React.

PR #1215, opened 2020-11-27
[FTAMS102020] Alejandro Pliakas Benitez

apliakas:


src/App.js
```diff
@@ -1,24 +1,52 @@
...
+
+  const [contacts, setContacts] = useState(contactsArray);
+
+  const addRandomContact = () => {
+    let copyContacts = [...contacts];
+    let randomContact = '';
+    
+    do {
+      randomContact = importedContacts[Math.floor(Math.random() * importedContacts.length)];
+    } while (contacts.some(actor => actor.id === randomContact.id));
```
wax-bear, line 19: Instead of this loop, you could simply splice the selected random contact from the importedContacts, and push that contact back again to the importedContacts upon deletion.

src/App.js
```diff
@@ -1,24 +1,52 @@
...
+
+    const newContacts = copyContacts.push(randomContact)
+
+    setContacts(copyContacts, newContacts);
```
wax-bear, line 23: Does this method need two arguments?

src/Contacts.js
```diff
@@ -0,0 +1,30 @@
...
+const Contacts = ({ contacts, randomContact, sortPopularity, sortName, deleteContact }) => {
+    
+  return (
+    <table>
+      <tr className='tableRow'>
+        <th><button onClick={randomContact}>Add Random Contact</button></th>
+        <th><button onClick={sortName}>Sort by Name</button></th>
+        <th><button onClick={sortPopularity} >Sort by Popularity</button></th>
+      </tr>
+      <tr className='tableRow title'>
```
wax-bear, line 16: You could store this `tr` in a `thead`.

src/Contacts.js
```diff
@@ -0,0 +1,30 @@
...
+        <th>Picture</th>
+        <th>Name</th>
+        <th>Popularity</th>
+        <th>Action</th>
+      </tr>
+      
+      {contacts.map(contact => <Contact deleteButton={deleteContact} contact={contact}/>)}
```
wax-bear, line 23: These `tr`s can be stored in a `tbody`.

src/App.js
```diff
@@ -1,24 +1,52 @@
-import React from 'react';
+import React, {useState} from 'react';
 import logo from './logo.svg';
```
wax-bear, line 2: This line results in an eslint error.

wax-bear (commented):
Great stuff Alex!
Please check my comments for potential improvements.

src/App.js
apliakas: but then how do I know that the random contact is not already in the contacts that are in the array?

wax-bear: If you splice contact X from the importedContacts, and push it to the contacts in the state, there is no X in importedContacts, and X in the state contacts. Upon deletion, contact X can be spliced from the state contacts, and be pushed to the importedContacts.

PR #1330, opened 2021-01-21
[Remote-PT-092020] Dirk Kelderman

dirkkelderman:


src/App.js
```diff
@@ -1,13 +1,75 @@
 import React from 'react';
 import './App.css';
+import contacts from './contacts.json';
+
+// const newContacts = contacts.slice(0,5)
+
 
 class App extends React.Component {
+
```
In-S4n3, line 9: Good job ma man! Everything working fine!

PR #1331, opened 2021-01-22
done

ZuzanaAK:


src/App.js
```diff
@@ -1,13 +1,133 @@
 import React from 'react';
 import './App.css';
-
+import contacts from './contacts.json';
+ 
 class App extends React.Component {
+  state = {
```
In-S4n3, line 6: Wooow, look at you all fancy with CSS! Gotta take a few lessons with you! Very good job!

PR #1334, opened 2021-01-26
Complete 1-4

MADFutcher:
:)


src/components/contacts/AddRandom.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react'
+import contacts from '../../data/contacts.json'
+
+class AddRandomContact extends React.Component{
```
In-S4n3, line 4: Very good job mate! You went one step ahead creating components! Everything working properly. 👍🏻

PR #1337, opened 2021-02-02
[Remote WD PT 280121] Thomas Schweers

t4ms:


t4ms:
Actually not sure if I pull the right repo or the codealongrepo

src/App.js
```diff
@@ -1,13 +1,76 @@
 import React from 'react';
 import './App.css';
+import contacts from './contacts.json';
 
-class App extends React.Component {
+export default class App extends React.Component {
+
```
In-S4n3, line 6: Very nice job Thomas, your style is always on point, as matter of fact I need lessons from you and I am not joking!! The code looks nice and clean.

PR #1512, opened 2021-07-09
[PT-AMS-MAR21] Mohamud Issa

MohamudIssa:
Learned a lot on react state hooks and passing down and lifting state.

and as always made plenty of mistakes along the way.


guido-an:
Well done Mohamud!

PR #1651, opened 2021-11-03
[FT-RMT-092021] Aintzane Morales

AintzaneM:


AintzaneM:
Hello!
Difficult lab.
I modified iteration 1 (just remembered "I can not rely on previous data"). 
On iteration 2 I used a push method to add a new actor on the initial array. Each time the bottom is clicked, it adds two times the same actor... I'm trying to figure out why.
I tried iteration 3 but I didn't managed to finished it yet. I will continue working on it during this last days of the week or weekend.

:) see you tomorrow!

PR #1659, opened 2021-11-07
[FT-RMT-092021] Geoffrey n°2

GeoLvl:
Lab Retry from scratch


GeoLvl:
Iteration 2 : problem with item key and 2 ppl appears instead of 1

PR #1723, opened 2022-02-23
WDFT Remote Feb 2022 - Stephan Geyer

stephan2311:


TA-Remote:
- Great work on the lab! I see you have tried to implement delete button. If you were to check documentation for splice() method, that's the return value:  `An array containing the deleted elements.`

So this won't work, as you are saying new state is equal to  array of deleted elements.

- Try applying filter() method here instead ;)

src/App.js
```diff
@@ -17,11 +17,23 @@ function App() {
...
 
+  const sortByPopularity = () => {
+    contacts.sort((a,b) => a.popularity < b.popularity ? 1 : -1);
+    setContacts([...contacts])
+  }
+
+  const sortByName = () => {
+    contacts.sort((a,b) => a.name.localeCompare(b.name));
+    setContacts([...contacts])
+  }
```
TA-Remote, line 28: Another thing to mention. As we cannot edit the state directly your syntax is not exactly right.

Sort() method mutates array, so you want to sort copied array, and only then pass the result into `setContacts()`

PR #1757, opened 2022-05-10
WDFT - RUBEN POELEN- done.

coderKron:


TA-Remote (commented):
Great work on the lab! Make sure you note the comments about copying array correctly.

Also for delete functinality, instead of adding another property to button, you can pass item's id as parameter into the delete function

```
<button onClick={ () => removeContact(x.id) }>Delete</button>
```

PR #1762, opened 2022-05-10
[FT WD REMOTE MARCH 22] - Thibaut Lartigue

thibautDevWorld:
Done iteration 3 !! Was really tedious


src/App.js
```diff
@@ -1,23 +1,56 @@
-import logo from './logo.svg';
+import { useState } from 'react';
 import './App.css';
+import contactsArr from './contacts.json';
 
 function App() {
+
+  const arrOfFive = contactsArr.slice(0, 5).map((elm) => {
+    return elm
+  })
```
TA-Remote, line 9: Slice() method returns a new array containing the extracted elements. So mapping is not needed here

src/App.js
```diff
@@ -44,6 +46,8 @@ function App() {
                 <td><img src={contact.pictureUrl} alt={contact.name}/></td>
                 <td>{contact.name}</td>
                 <td>{contact.popularity}</td>
+                <td>{ contact.wonOscar === true ? <span className='trophee'>🏆</span> : ""}</td>
+                <td>{contact.wonEmmy === true ? <span className='trophee'>🏆</span> : "" }</td>
```
TA-Remote, line 61: bellow two lines are equivalent, as they both return true. 

`if (contact.wonOscar === true) `
`if (contact.wonOscar)`

Remember that inside if condition (or in this case in ternary expression condition), we get true or false.

So when we say:
` if ('hello') `-->it will return true

Below gives falsy:
`if ('')` -> returns false
`if (false) `-> retruns false

https://developer.mozilla.org/en-US/docs/Glossary/Falsy

src/App.js
```diff
@@ -11,13 +11,21 @@ function App() {
 
  const [contacts, setContact] = useState(arrOfFive);
 
+ 
 
+  const addRandomContact = () => {
+    
+    let filterList = contactsArr.filter((newContact) => {
+      return !contacts.includes(newContact)
+      })
```
TA-Remote, line 20: Always make a copy of the array. Get into habit of working on copied data rather than data you have received from the database, or original source. 

You can simply do
```
let filterList = [...contactsArr].filter
```

src/App.js
```diff
@@ -11,13 +11,21 @@ function App() {
...
 
+    const randomItm = filterList[Math.floor(Math.random()* filterList.length)]
 
+    setContact(contacts.concat(randomItm))
```
TA-Remote, line 24: We do not want to edit original array -> even if it's adding values.
Instead, you want to grab the previous/latest value and update it:

```
setContacts(contacts => [...contacts, randomItm])
```

TA-Remote (commented):
Great work on the lab Thibaut. 

All the labs are quite important now to make sure you practice-understand fundamentals, so always make the most effort you can -> without burning yourself out ;) You can review solutions once we upload them, and see if you understand it

PR #1765, opened 2022-05-10
[WDFT MAR2022] Ivo Silva - Completed Iterations 1 - 5

Ivo-S-Silva:


TA-Remote:
Amazing, Ivo! :)

TA-Remote:
just one thing: because we work with previous values in our state, we should use the second syntax (option 2 that Luis showed you) with callback function and return -> for setState function

PR #1766, opened 2022-05-10
WDFTMarch2022Remote Sepid

Sepidehatt:
done!


src/App.js
```diff
@@ -3,19 +3,21 @@ import './App.css';
 import { useState } from 'react'
 
 function App() {
-  const fiveContact = contactsArr.splice(0, 5);
-
+  const fiveContact = contactsArr.filter((contact,index)=> index<5);
+console.log(fiveContact)
```
TA-Remote, line 7: Alternative solution:

```
const fiveContacts = contactsArr.slice(0, 5);
```

src/App.js
```diff
@@ -3,19 +3,21 @@ import './App.css';
...
   const addRandom = () => {
-    const randomContact = otherContacts.slice(randomIndex,randomIndex +1)
-    const addToContacts = shallowCopyContact.concat(randomContact)
-    console.log(addToContacts)
+    // const randomContact = otherContacts.splice(randomIndex,1)
+    const addToContacts = shallowCopyContact.concat(otherContacts[randomIndex])
+    // const addToContacts = shallowCopyContact.concat(randomContact)
+    // console.log(addToContacts)
     setContact(addToContacts)
   }
```
TA-Remote, line 22: To keep code more readable, I would suggest keeping all logic in separate blocks of code/separate functions.

So here, if you have a function to addRandom -> you might want to move lines 10,11  inside it.  Not to have loose lines of code

TA-Remote (commented):
Amazing Sepid! Keep working hard

src/App.js
Sepidehatt: i used it before,,,but i came across lots of problem,,, then i changed it,,

src/App.js
Sepidehatt: cool..thank you for your suggestion

PR #1768, opened 2022-05-10
WDFT Remote Mar22-Ha Thu TRAN

HaThu6989:


TA-Remote:
Hey Thu, good job, but I found some things to improve: you used setState without callback and return and addRandomContact adds random contact but it's always same 5 contacts + one random (never more than 6 contacts). More in detail in following comments :)

src/App.js
```diff
@@ -1,25 +1,83 @@
...
+  let fiveContacts = contactsArr.slice(0, 5);
+
+  /** Iteration 3 | Add New Random Contacts **/
+  const [contacts, setContact] = useState(fiveContacts);
+  const remainingContact = contactsArr.slice(6, contactsArr.length);
+  const randomContactIndex = Math.floor(
+    Math.random() * remainingContact.length
+  );
+  const randomContact = remainingContact[randomContactIndex];
+  const randomArr = fiveContacts.concat(randomContact);
```
TA-Remote, line 15: The issue with those five/six contacts should be solved when you do 
`  const randomArr = [...contacts].concat(randomContact);`
Because you always use randomArr and randomArr is always same five contacts (fiveContacts) and not your state. 
If you use a copy of state here instead you can have more that 6 contacts and not always same 5. 
Let me (Joanne) know if you have any questions

TA-Remote: Also: it would be better if you use setState with callback

src/App.js
```diff
@@ -1,25 +1,83 @@
...
+  // console.log(randomArr);
+
+  function addRandomContact() {
+    setContact(randomArr);
+  }
```
TA-Remote, line 20: an example would be this solution:
```
 const addRandomContact = ()=>{
    setContacts((prevContacts)=>{
      const newContactsList = contactsArr.filter((contact)=> {
        return !prevContacts.includes(contact);
      });
      const getRandomContact = newContactsList[Math.floor(Math.random()* newContactsList.length)];
      const newConctactSelection = [...prevContacts,getRandomContact]
      return newConctactSelection;
    })
  };
```

src/App.js
```diff
@@ -1,25 +1,83 @@
...
+
+  /** Iteration 4 | Sort Contacts by Name and Popularity **/
+  function sortByPopularity() {
+    const orderByPopularity = randomArr.sort(
+      (a, b) => b.popularity - a.popularity
+    );
+    setContact(orderByPopularity);
+  }
```
TA-Remote, line 28: same here. 
example:
```
const sortByPopularity = () =>{
    setContacts((prevContacts)=>{
      let sortedByPopularity = [...prevContacts].sort((a, b) => b.popularity - (a.popularity))
      return sortedByPopularity
    })
  }
```

src/App.js
```diff
@@ -1,25 +1,83 @@
...
+
+  function sortByName() {
+    const orderByName = randomArr.sort((a, b) => a.name.localeCompare(b.name));
+    setContact(orderByName);
+  }
+
```
TA-Remote, line 34: and for the last functions the same. 
But the logic itself is correct :)

PR #1769, opened 2022-05-10
WDFT MAR22 Harry

hmarchbank:
1. that lab was just plain evil
2. just about did all of the iterations at some points via trial and error it felt like.
3. only realised my iteration 2 wasn't working as promised towards the end, too dead to go back.
4. I think i kinda was a bit cheeky with the delete function and just went for straight DOM manipulation but it works, what was the way I was supposed to accomplish it?
5. Think it would be helpful to go over states again in class, what they return was just plain confusing.


TA-Remote:
Hey Harry, your message suggests worse, you didn't do that bad, but probably without understanding everything. I'll try to list a few things, otherwise I hope that today gives you a little more understanding for state.

src/App.js
```diff
@@ -1,23 +1,74 @@
...
 
+
+
 function App() {
+  let fiveContacts =  [...contactsArr.slice(0, 5)]
+  
+  let [contactList, setContacts] = useState(fiveContacts)
+
+  const addContact = ( () => {
+    let random = Math.floor( (Math.random() * contactsArr.length) -1)
```
TA-Remote, line 13: here you forget the logic for adding contacts once. We have already selected 5 contacts from contactsArr and should avoid selecting these actors again

src/App.js
```diff
@@ -1,23 +1,74 @@
...
+        if (index === random){
+          return true
+        } else {
+          return false
+        }
+      })
+      return contactList.concat(newArr)
+    })
+
+  })
```
TA-Remote, line 26: I tried to explain that this morning, so for example like this would be a good solution:
```
const addRandomContact = ()=>{
    setContacts((prevContacts)=>{
      const newContactsList = contactsArr.filter((contact)=> {
        return !prevContacts.includes(contact);
      });
      const getRandomContact = newContactsList[Math.floor(Math.random()* newContactsList.length)];
      const newConctactSelection = [...prevContacts,getRandomContact]
      return newConctactSelection;
    })
  };
```
let me know if you have any questions

src/App.js
```diff
@@ -1,23 +1,74 @@
...
+  
+  const sortByName = ( () => {
+    const orderName = [...contactList.sort((a, b) => a.name.localeCompare(b.name))]
+    setContacts(orderName)
```
TA-Remote, line 30: this one is great, it's just that you should use the setState syntax with callback and return, for example like this:
```
const sortByName = ()=>{
    setContacts((prevContacts)=>{
      let sortedContacts = [...prevContacts.sort((a, b) => a.name.localeCompare(b.name))]
      return sortedContacts
    })
  };
```

src/App.js
```diff
@@ -1,23 +1,74 @@
...
+  })
+  const sortByPop = ( () => {
+    const orderName = [...contactList.sort((a, b) => a.popularity - b.popularity)]
+    setContacts(orderName)
+  })
```
TA-Remote, line 35: same here

src/App.js
```diff
@@ -1,23 +1,74 @@
...
+
+  const deleteRow = ((el) => {
+    const target = document.getElementById(el.target.id)
+    target.parentNode.removeChild(target)
+    setContacts(contactList)
+  })
```
TA-Remote, line 41: react is doing something like dom manipulation, so we should avoid to use dom manipulation with react.
what we do instead is passing the contact id to the function inside our map method. So when we map over contacts, we display contact.name and all that stuff, so we can use contact.id (if you check json data, we have an id) and we pass that id to the delete function as an argument and our function can look like this:
```
 const deleteContact = (contactId) =>{
    setContacts((prevContacts)=>{
      const newContactsList = prevContacts.filter((contact)=> {
        return contact.id !== contactId;
      });
      return newContactsList
    })
  }
```

inside return/ inside map() :
    `            <td> <button onClick={ () => deleteContact(contact.id) }>Delete</button> </td>`

PR #1771, opened 2022-05-10
WDFT MArch2022 Valentin Paul

Valentin-Paul:
only solved a few iterations.
underestimated the time I need and had to go to sleep in the end to survive the next day.


src/App.js
```diff
@@ -1,23 +1,66 @@
...
+   }
+
+  const sortByAlphabet = ()=>{
+    setContacts(prevContacts=>{
+      const alphabeticList = prevContacts.sort((a,b)=>{
+        return a.name.localeCompare(b.name)
+      })
+      return alphabeticList;
+    })
+  }
```
TA-Remote, line 26: For this iteration, you firstly want to create a copy of the array you'll be working on, you do not really want to touch the original array of data - not to corrupt it. 

Otherwise you have correct solution here! 

For example:

```
const sortByAlphabet = ()=>{
    setContacts(prevContacts=>{
      const alphabeticList = [...prevContacts].sort((a,b)=>{
        return a.name.localeCompare(b.name)
      })
      return alphabeticList;
    })
  }
```

TA-Remote (commented):
Great work on the lab Valentin! 
Look at my comments and let me know if some points still not clear. All the labs are quite important now to make sure you practice-understand fundamentals :)

PR #1773, opened 2022-05-11
WDFT Remote Mar22 Alexander Nerz

Alex-WD-22:
For iteration 3 I had some help from Eduardo. Iterations 4,5,6 I started but didn't finish yet.


src/App.js
```diff
@@ -1,23 +1,72 @@
...
+
+    let randomIndex = otherContacts[Math.floor(Math.random() * otherContacts.length)]
+    setContacts(contacts.concat(randomIndex))
+  }
+
+  let sortPopularity = () => {
+   contacts.sort((a, b) => (a.popularity < b.popularity) ? 1 : -1)
+    setContacts(sortPopularity)
+  }
+
```
TA-Remote, line 28: Here you cannot call your function as argument in your function. 
Better make copy of your contacts before sort, and then setContacts with this new sortedContacts:

```
const sortByPopularity = () => {
  const sortedContacts = [...contacts.sort((a,b) => (a.popularity < b.popularity) ? 1 : -1)];

  setContacts(sortedContacts);
};
```

src/App.js
```diff
@@ -1,23 +1,72 @@
...
 import './App.css';
 
+
+
 function App() {
+  let fiveContacts = contactArr.slice(0, 5)
+
+  const [contacts, setContacts] = useState(fiveContacts);
+
+  let otherContacts = contactArr.slice(5, contactArr.length)
```
TA-Remote, line 12: no need to slice the rest contactsArr before filter, since contacts in your state only holds the ones that are displaying already

TA-Remote:
Hi Alexander, your approaches were good. Look into labs solution to proceed further and understand concepts better

PR #1774, opened 2022-05-11
it 3 done, 4 & 5 remaining ALEJANDRO BLANCO 2022

abg1995:


TA-Remote:
Iteration 1-3 well done! Just some small suggestions and hints for Iteration 4 and 5:

**Iteration 1:**
to slice first 5 contacts -> .slice(0, 5) -> 0 is the first index and 5 is the number of items you want to slice

**Iteration 2:** 
in your function oscar & emmy better NOT use the word “props”  as argument - in react “props” is reserved to lift up state (just a convention to stick to, and not confusing)

**Iteration 3:** 
good

**Iteration 4: (hint)**
function sortByName  --> use .sort() method and localeCompare and call setContacts funtion with sortedContacts 

function sortByPopularity --> use .sort() method  and call setContacts funtion with sortedContacts  + when render use Math.round() 

**Iteration 5: (hint)**
function removeContact --> use .filter() method to check for id and call setContacts function with filteredContacts
Add button with onClick to call that function

PR #1775, opened 2022-05-11
Iteration 1-3

didicozaur:


src/App.js
```diff
@@ -1,23 +1,61 @@
...
+          <tr>
+            <th>Picture</th>
+            <th>Name</th>
+            <th>Popularity</th>
+            <th>Oscar</th>
+            <th>Emmy</th>
+          </tr>
+        </thead>
+        {fiveContacts.map((contact) => {
+          return (
```
TA-Remote, line 48: You don't map only through your fiveContacts, but through contacts, that will change dynamically, thanks to 
-- > const [contacts, setContacts] = useState(fiveContacts)

"fiveContacts" is only your initial value
e.g. if you add a random contact, then your setContacts function will change again. 

```
 {contacts.map((contact) => .... )}
 
```

PR #1778, opened 2022-05-24
[FTLIS042022] Francisco Bento

FranciscoGBento:


src/App.js
```diff
@@ -1,23 +1,71 @@
...
-import './App.css';
+import contacts from "../src/contacts.json";
+import React, { useState } from "react";
+import "./App.css";
 
 function App() {
+  const [fiveContacts, setFiveContacts] = useState(contacts.slice(0, 5));
+
+  const addContact = (contactAdded) => {
+    const newContact =
```
webLisbonTA, line 9: A good idea here is to apply Math.random only on the remaining contacts on the list, this way you can get repeated contacts.
Also it is important to get the length of the whole array, not just 5

webLisbonTA (commented):

PR #1780, opened 2022-05-24
[FTLIS042022] João Gonçalves

joaopdg:
lab done


src/App.js
```diff
@@ -1,23 +1,72 @@
-import logo from './logo.svg';
-import './App.css';
+import "./App.css";
+import React, { useState } from "react";
+import contactsData from "./contacts.json";
 
 function App() {
+  const [contacts, setContacts] = useState(contactsData.slice(0, 5));
+
+  const adRandom = () => {
```
webLisbonTA, line 8: A good idea here is to apply Math.random only on the remaining contacts on the list, this way you can get repeated contacts

webLisbonTA (commented):

PR #1781, opened 2022-05-24
[WDFT042022] Dino Marchiori

dinaco:


src/App.js
```diff
@@ -1,23 +1,49 @@
...
-import './App.css';
+import "./App.css";
+import React, { useState } from "react";
+import contacts from "./contacts.json";
+import Contacts from "./components/Contacts";
 
 function App() {
+  const [contact, setContact] = useState(contacts.slice(0, 5));
+  const createNew = () => {
+    const newIndex = Math.floor(Math.random() * contacts.length);
```
webLisbonTA, line 9: A good idea here is to apply Math.random only on the remaining contacts on the list, this way you can get repeated contacts

webLisbonTA (commented):
Well done!

PR #1784, opened 2022-05-24
[FTLIS042022] matildeCosta

matildedacosta:


src/App.js
```diff
@@ -1,23 +1,101 @@
...
-import './App.css';
+import React, { useState } from "react";
+import contacts from "./contacts.json";
+import "./App.css";
 
 function App() {
+  const [allContacts, setAllContacts] = useState(contacts.slice(0, 5));
+
+  const randomizeMovie = () => {
+    const randomMovie = contacts[Math.floor(Math.random() * contacts.length)];
```
webLisbonTA, line 9: A good idea here is to apply Math.random only on the remaining contacts on the list.

webLisbonTA (commented):

PR #1785, opened 2022-05-24
[FTLIS042022] Mariana Oliveira

miloliveira:
lab done


src/App.js
```diff
@@ -1,23 +1,73 @@
+import {useState} from 'react';
 import logo from './logo.svg';
 import './App.css';
+import contactData from "../src/contacts.json"
 
 function App() {
+
+  const [contacts, setcontacts] = useState(contactData.slice(0,5))
+  
+  const randomMovie = ()=>{
```
webLisbonTA, line 10: A good idea here is to apply Math.random only on the remaining contacts on the list.

webLisbonTA (commented):

PR #1786, opened 2022-05-24
[WD_APR_2022] João Morgado

JoaoMorgado8:


src/App.js
```diff
@@ -1,23 +1,91 @@
...
-import './App.css';
+import "./App.css";
+import { useState } from "react";
+import contacts from "../src/contacts.json";
 
 function App() {
+  const [contact, setContact] = useState(contacts.slice(0, 5));
+
+  const randomContact = () => {
+    let random = contacts[Math.floor(Math.random() * contacts.length)];
```
webLisbonTA, line 9: A good idea here is to apply Math.random only on the remaining contacts on the list, this way you can get repeated contacts

webLisbonTA (commented):

PR #1788, opened 2022-05-24
LISFTWD Marcelo Milhomem

marcelomilhomem:


src/App.js
```diff
@@ -1,23 +1,75 @@
...
+import "./App.css";
+import contacts from "../src/contacts.json";
+import React, { useState } from "react";
 
 function App() {
+  const [contactsList, setContactsList] = useState(contacts.slice(0, 5));
+
+  const addRandom = () => {
+    const random = contacts[Math.floor(Math.random() * contacts.length)];
+    setContactsList([random, ...contactsList]);
```
webLisbonTA, line 10: A good idea here is to apply Math.random only on the remaining contacts on the list.

webLisbonTA (commented):

PR #1790, opened 2022-05-24
[WDFT0422]Duarte Alves

drttyy:


src/App.js
```diff
@@ -1,23 +1,90 @@
-import logo from './logo.svg';
-import './App.css';
-
+import "./App.css";
+import { useState } from "react";
+import ContactData from "./contacts.json";
 function App() {
+  const [contacts, setContacts] = useState(ContactData.slice(0, 5));
+
+  const randomContacts = () => {
```
webLisbonTA, line 7: A good idea here is to apply Math.random only on the remaining contacts on the list, this way you can get repeated contacts

webLisbonTA (commented):

PR #1791, opened 2022-05-24
[WDFT042022] João Antunes

joaocbantunes:
lab done


src/App.js
```diff
@@ -1,25 +1,79 @@
...
+import { useState } from "react";
+
+const fiveContacts = contacts.slice(0, 5);
 
 function App() {
+  const [contact, setContact] = useState(fiveContacts);
+
+  const newContacts = contacts.slice(6);
+
+  const randomContact = () => {
```
webLisbonTA, line 12: If you want to get a random contact that is not listed, you should apply the Math random on newContacts array, not on the original one.

webLisbonTA (commented):

PR #1794, opened 2022-05-24
FTWD0422 CATARINA SALVADO

catarinapsalvado:


src/App.js
```diff
@@ -1,25 +1,76 @@
...
 
 function App() {
+  const [contactList, setContactList] = useState(contacts.slice(0, 5))
+  const [randomContact, setRandomContact] = useState(contacts.slice(6))
+
+
+ 
+
+  const newContact = () => { 
+    const newContact = contacts[Math.floor(Math.random()* contacts.length)]
```
webLisbonTA, line 14: If you want to get a random contact that is not listed, you should apply the Math random on randomContact array, not on the original one.

webLisbonTA (commented):

PR #1795, opened 2022-05-24
[WDFT0422] Marina Palma

MarinaPalma:


src/App.js
```diff
@@ -1,23 +1,87 @@
-import logo from './logo.svg';
+
 import './App.css';
+import React, {useState} from 'react';
+import contactData  from "./contacts.json"
 
 function App() {
+const [contacts, setContacts] = useState(contactData.slice(0,5));
+const randomContact = () =>{
```
webLisbonTA, line 8: A good idea here is to apply Math.random only on the remaining contacts on the list.

webLisbonTA (commented):

PR #1804, opened 2022-07-26
[RMFT June 2022] Alex's Lab

Virgule09:


Virgule09:
I took a break, ate diner and finally everything clicked. I was using the useState hook multiple times when I was supposed to reuse the function.
Tried it with the delete functionality and it worked. From then on it was just a formality.
During the whole day I had no difficulties understanding the concept so this was very frustrating, I am happy that I managed to get it done, finally ! :)
PS: I believe my table is not working, or maybe it is just CSS, but I must confess that I did not care too much about this. But I already know that I could have done better there, sorry about that.

src/App.js
```diff
@@ -1,25 +1,62 @@
...
 function App() {
+
+  const [contactsArr, setContact] = useState(contacts.slice(0, 5));
+  const addRandomContact = () => {
+    const randomContact = contacts[Math.floor(Math.random() * contacts.length)];
+    setContact((prevContacts) => {
+      const newContacts = [...prevContacts, randomContact]
+      return newContacts
+    })
+  }
```
TA-Remote, line 14: Last step for this lab would be to display list without repeated contacts :) 

Clue: you might want to create new array that removes contacts that already being displayed, and get random contact from that new list

TA-Remote (commented):
The most important part is that you didn't give up 😄  and it clicked eventually, well done!

src/App.js
Virgule09: Right !!! I completely forgot about this :/

PR #1810, opened 2022-07-26
[RMFT June '22] Lisa Hilterhaus: lab done

cortinaHs:


src/components/ContactList.js
```diff
@@ -0,0 +1,102 @@
...
+        setContacts((prevList) => {
+            const randomContact = contactData[Math.floor(Math.random()*contactData.length)];
+            const existingContact = prevList.find(element => element.id === randomContact.id)
+            if(!existingContact){
+                return prevList.concat(randomContact)
+            } else {
+                addRandomContact();
+            }
+        })
+    }
```
TA-Remote, line 22: Yuu have that bug we discussed in the morning, that app crashes if we click 'add' many times.

One solution would be to use `filter() `instead of` find()`. 
So you want to first create an array where you take list from database and remove all the contacts already displayed. And get random contact from it instead

```
    const addRandomContact = () => {
        setContacts((prevList) => {
            const remainingContacts = contactData.filter((contact)=> {
                return !prevList.includes(contact);
              });

            const randomContact = remainingContacts[Math.floor(Math.random()*remainingContacts.length)];
            return prevList.concat(randomContact)
        })
    }
```

TA-Remote (commented):
Great job!

PR #1813, opened 2022-07-27
not completed

Khalid9180:


src/App.js
```diff
@@ -1,25 +1,61 @@
...
+      <table>
+        <tr>
+          <th>Picture</th>
+          <th>Name</th>
+          <th>Popularity</th>
+          <th>WonOscar</th>
+          <th>WonEmmy</th>
+        </tr>
+        </table>
+      {newArr.map((element) => {
```
TA-Remote, line 34: here instead of newArr, you need to put contacts here, because contacts is our stateful variable and when you wanna add new rando file, it should change ;)
contacts.map()

Khalid9180:
hmmm

Khalid9180:
Is that the reason why the button was not working ?

TA-Remote:
yes :)

PR #1814, opened 2022-07-27
Nic Solitom - incomplete, only started iteration 3.

nicsolitom:
Due to only starting with the labs so late after a long day of following/coding along, I did not have more energy or focus in time for this lab.


src/App.js
```diff
@@ -1,23 +1,76 @@
...
+
+
+/// NOT READY!
+  function addRandomContact() {
+    let randomNum = Math.floor(Math.random() * contactsArr.length);   
+    setContacts((prevState) => {
+      let returnArr = [];
+      return prevState.push(contactsArr[randomNum]);
+    });
+
```
TA-Remote, line 36: Just one note here, never write functions inside functions like this.

Instead follow this syntax
`const addRandomContact = () => {}`

TA-Remote (commented):
As you said yourslf, if you'll find time work on this lab further! alternatively the solution is uploaded on student portal, so you can analyse it :)

PR #1818, opened 2022-07-27
Jack Le

jackletran:
iteration 1, 2, 5 - tried 3 - didn't finish 4


src/App.js
```diff
@@ -1,25 +1,77 @@
-import logo from './logo.svg';
-import './App.css';
+import "./App.css";
+import contactsArray from "./contacts.json";
+import { useState } from "react";
 
 function App() {
+  const contacts = [
```
TA-Remote, line 6: you can use method here to catch first five contact

src/App.js
```diff
@@ -1,25 +1,77 @@
...
-        >
-          Learn React
-        </a>
-      </header>
+      <button
+        onClick={() => {
+          setActors((prevActors) => {
+            let randomIndex = Math.floor(Math.random() * contactsArray.length); // random index
+            let randomContact = contactsArray[randomIndex]; // pick a random from whole list
+            prevActors.push(randomContact); // push to newList
```
TA-Remote, line 31: you can use push method whenever you have shallow copy, because it modifies directly your array,
here you can use this syntax to update your state :
 const newArr = [...prevActors , randomContact]
and return newArr

PR #2105, opened 2023-04-04
[RMT WD-FT FEB-2023] Ismael Freitas

IsmaelFreitas98:
I did not implement styling. Instead I allowed the deleted contacts to be sent back to the available list, so it would make it possible to call them again 😄


ykztenoks:
Hey Ismael!
Congrats on your lab, awesome job on your code with great usage of `useState()` and conditionals.
One improvement would be for the `sortByName()` function, you could use `localeCompare()` for cleaner code and better sort.

PR #2115, opened 2023-04-04
[RMT.WD.FT.FEB-23] Sotiris Ntokas

SotDok:
Had problem with iteration 3.


ykztenoks:
Hey Sotiris!
Congratulations on your lab, clean and concise code. As always if you have time make sure to clear up the code avoiding unnecessary comments and extra spaces.
For the iteration #3 you could've used the `setContacts()` to set the state of the list of contacts, using the spread operator to keep the previous contacts and using the randomNumber to get an element in the contacts.json array, just like:
```  const addRandomContact = () => {
    const randomContact = otherContacts[Math.floor(Math.random() * otherContacts.length)];
    if(!contacts.includes(otherContacts[randomContact)){
       setContacts([...contacts, otherContacts[randomContact])
  } 
  ```
  
  I'm first checking if the contacts state has the random element of the remaining contacts so it doesn't duplicate, as for that you have a mistake as you don't have the remaining contacts in the otherContacts state, instead of:
   `  const [otherContacts, setOtherContacts ] = useState(contactsFromJson.slice(0, 5));`
   you should:
   `  const [otherContacts, setOtherContacts ] = useState(contactsFromJson.slice(5);`

Because in your contacts state you have the first five, and for the second one you'd need the remaining, so you slice from the end of your first slice with no need for a second parameter.

PR #2116, opened 2023-04-04
[WDRMTFT20.02.2023]Peng Chen

Mr-Chennn:
done

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,16725 @@
+{
+  "name": "lab-react-ironcontacts",
+  "version": "1.1.0",
+  "lockfileVersion": 3,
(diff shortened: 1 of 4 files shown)


ykztenoks:
Hey Peng!
Awesome work on your lab, clean and concise code.
Few improvements would be for having a second state storing the remaining contacts, such as:
`const [remaining, setRemaining] = useState(data.slice(5))`
In the addRandomContacts() function you're missing a conditional statement to avoid duplicate contacts, having that remaining state would prevent that for you.
As for sorting by names you could use the `localeCompare()` for a cleaner code and safer sort.
Other than that, keep it up!

PR #2119, opened 2023-04-04
[RMT FT Feb2023] Bartosz Dubinski

bartoszde:
It was challenging lab. I spend the most time with iteration 3 and 4. I did simple css, just because I wanna feel I can do something without problems. 

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,16725 @@
+{
+  "name": "lab-react-ironcontacts",
+  "version": "1.1.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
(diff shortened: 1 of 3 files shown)


TA-Remote:
Hey Bartosz, great job!
 Of course iteration 3 and 4 were the trickiest 😉 just small suggestions: 
- When you wanna sort alphabetically, good practice is o use .localeCompare() —>
```
const sortedByName = toSortName.sort((a, b) =>
      a.name.localeCompare(b.name) // A-Z
    );
```
- For the addRandom we need to keep in mind that we don’t want to repeat contacts, good practice would be to create another state variable to store the remaining contacts that are not displayed yet. (when I completed this lab during my corse I forgot it too😜)
Love your memes

PR #2120, opened 2023-04-04
[rmt-wd-ft-0223] Pim

8lom:
iteration 5 done, iterations 3,4blocked.
Got a lot of work to do

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,16725 @@
+{
+  "name": "lab-react-ironcontacts",
+  "version": "1.1.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "lab-react-ironcontacts",
+      "version": "1.1.0",
+      "dependencies": {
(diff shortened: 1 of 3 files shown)


ykztenoks:
Hey Pim!
Great work on your development
For the iteration #3 you could've used the setContacts() to set the state of the list of contacts, using the spread operator to keep the previous contacts and using the randomNumber to get an element in the contacts.json array, just like:

    const randomContact = otherContacts[Math.floor(Math.random() * otherContacts.length)];
    if(!contacts.includes(otherContacts[randomContact)){
       setContacts([...contacts, otherContacts[randomContact])
  } 
I'm first checking if the contacts state has the random element of the remaining contacts so it doesn't duplicate, so for that you'd also need an extra state called otherContacts with the initial state set as:
 ` const [otherContacts, setOtherContacts ] = useState(contactsFromJson.slice(5);`
so you'd have also access to the rest of the contacts from the 5th position.

Other than that, keep it up!

PR #2122, opened 2023-04-04
[RMT-FT-FEB-2023] Paulius Mockevicius

pmockevicius:
Nice lab


ykztenoks:
Hey Paulius!
Great work on your lab! Great usage of states and conditionals.
For the `sortByName()` function, you could've used `localeCompare()` for a cleaner code and safer sort.
Other than that keep up the great work!

PR #2124, opened 2023-04-05
[RMT FT 20 FEB 23]Hristos Pappas

hristosppp:


ykztenoks:
Hey Hristos!
Great work on your lab!
One improvement would be for the sortByName() function, you could use localeCompare() for cleaner code and better sort.

PR #2125, opened 2023-04-05
[RMT FT 20 FEB 23]Danial Raja

danialraja1600:
Completed all iterations


ykztenoks:
Hey Danial!
Great work on your lab.
One improvement would be for the sortByName() function, you could use localeCompare() for cleaner code and better sort.

PR #2190, opened 2023-06-19
 [FTRMT052023] Philippe Miossec

pmiossec:
🎥⭐
HollywoodSignXanaduGIF


TA-Remote (commented):
- Awesome work
- Good guards to avoid the app breaking
- Thanks for the GIF

PR #2198, opened 2023-06-20
[FTRMT052023] Pierre 

Dokryptos:
done

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -15,21 +15,39 @@ function App() {
 
 
   const addRandomContact = () => {  
-    console.log( randomArr);
+
     const intRandom = Math.floor (Math.random() * randomArr.length);
     const addContact = randomArr[intRandom];
 
-    setContacts(([...contactInfo, addContact]));
+    setContacts((contactActually => [...contactActually, addContact]));
+  }
+
+  const sortByName = () =>{
+    console.log(contactInfo);
+    const compareByAlphabetic = [...contactInfo].sort((a, b) => a.name.localeCompare(b.name))
+    setContacts(compareByAlphabetic)
   }
 
 
+  const sortByPopularity = () => {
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -1,25 +1,91 @@
...
-          Learn React
-        </a>
-      </header>
-    </div>
-  );
+  
+
+  const [contactInfo, setContacts] = useState(contacts.slice(0,5))
+
+  const randomArr = [...contacts].filter(contact => !contactInfo.includes(contact))
```
TA-Remote, line 11: I would write that piece of code inside of the `addRandomContact()` as you probably only use that one inside that function

src/App.js
```diff
@@ -1,25 +1,91 @@
...
+
+  
+  console.log(randomArr);
+
+
+  const addRandomContact = () => {  
+
+    const intRandom = Math.floor (Math.random() * randomArr.length);
+    const addContact = randomArr[intRandom];
```
TA-Remote, line 20: Also I would add a `if` statement to guard the function when you already added everyone, otherwise the application will break

src/App.js
```diff
@@ -1,25 +1,91 @@
...
+                        <td>{Math.floor( celebrity.popularity * 100 )/100}</td> 
+                        {celebrity.wonOscar
+                        ? <td>🏆</td>
+                        : <td></td>}
+                        
+                        {celebrity.wonEmmy
+                        ? <td>🏆</td>
+                        : <td></td>}
+
+                        <button onClick={() => {deleteContact(celebrity.id)}}>Delete</button>
```
TA-Remote, line 78: That one should probably inside a `<td>` so you get rid of the warning in the console of the browser

TA-Remote (commented):
- Good work

PR #2199, opened 2023-06-20
[FTRMT052023] eirikKlebo 

eiriktobias:


TA-Remote:
hey @eiriktobias 
well done with your iterations🙌

the addRandom was a bit tricky, but you got the rigth logic!
- store the remaining in a state variable 
` 
const [allContacts, setAllContacts] = useState(contacts); // --> contact.slice(5)  will give you all but the first 5
`
- update the remaining removing the random picked: 
  make a shallow copy first, then you can use different methods to remove the random element (filter, splice)
- update the state with the 'clean up' array using setAllContacts

keep it up💪
Alessandra

PR #2200, opened 2023-06-20
[FTRMT 08052023] Solid Zendrato

zendratodaeli:


src/App.js
```diff
@@ -1,23 +1,71 @@
...
+import { useState } from 'react';
 import './App.css';
+import contacts from "./contacts.json";
 
 function App() {
+
+  const firstFive = contacts.slice(0, 5);
+  const[contactsArray, setContactsArray] = useState(firstFive)
+
+  const randomizeContacts = () => {
```
TA-Remote, line 11: The `addRandom()` should add a random contact to the displayed contacts list, the logic to randomize the contacts work => But if the array is bigger or you need a real random array you should use another algorithm to do that one

src/App.js
```diff
@@ -1,23 +1,71 @@
...
+          <th>Picture</th>
+          <th>Name</th>
+          <th>Popularity</th>
+          <th>Won Oscar</th>
+          <th>Won Emmy</th>
+          <th>Actions</th>
+        </tr>
+      </table>
+      
+      {contactsArray.map(element => {
```
TA-Remote, line 55: You probably just wanted to add a row to the previous table, and also, remember that when we map through a list we should add a key that would be a unique identifier for the JSX we are rendering

TA-Remote (commented):
- Good work done :)

PR #2202, opened 2023-06-20
[FTRMT052023] Diogo Martins

disaacjm:
Iteration 1,2 and 3 done. Not able to do all. Too tired to think well about it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,16927 @@
+{
+  "name": "lab-react-ironcontacts",
+  "version": "1.1.0",
+  "lockfileVersion": 3,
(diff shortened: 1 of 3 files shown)


TA-Remote:
Hey @disaacjm 
Well done with your iterations, you got the energy to solve the trickiest one 🙌!

Small little tip: when declaring your state variable, name the two arguments following the convention (also mentioned in [react docs](https://react.dev/reference/react/useState) ) --> [ banana, setBanana ]

You'll have a lot of possibility to practice those concepts, and you are gonna use them after the bootcamp too!

Keep it up💪

Alessandra

PR #2206, opened 2023-06-20
[FTRMT052023] Viktor N.

ViktorNstr:
Done.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,23 +1,73 @@
-import logo from './logo.svg';
-import './App.css';
+import "./App.css";
+import contactsData from "./contacts.json";
+import { useState } from "react";
 
 function App() {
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -1,23 +1,73 @@
-import logo from './logo.svg';
-import './App.css';
+import "./App.css";
+import contactsData from "./contacts.json";
+import { useState } from "react";
 
 function App() {
+  const [contacts, setContacts] = useState(contactsData.slice(0, 5)); 
+
+  const addRandomContact = () => {
```
TA-Remote, line 8: For that one maybe you could limit if you don't have anymore contacts to add, because otherwise this breaks your application

TA-Remote (commented):
- Good work

PR #2208, opened 2023-06-20
[FTRMT052023]

jfaria23:
I needed a lot of help from my web dev sister on iteration 3 and 4. 
Adding random contact is not working beauce of : 
Cannot read properties of undefined (reading 'pictureUrl')
TypeError: Cannot read properties of undefined (reading 'pictureUrl')

Even with help I don't fully understand much of this code.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,16927 @@
+{
+  "name": "lab-react-ironcontacts",
+  "version": "1.1.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "lab-react-ironcontacts",
+      "version": "1.1.0",
+      "dependencies": {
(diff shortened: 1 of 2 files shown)


TA-Remote:
hey @jfaria23 
thanks for your honesty!! 🙂
 - about the error: when you are generating a random num you should use Math.floor( Math.random .......), without it the number is not an integer, so your randomCelebrity is 'undefined', and .map trows an error when try to read 'undefined' elements.... the error refers to pictureUrl just because is the first property it's trying to read .
 - if you still have doubs about array methods(like .sort), that you'll be using a loooooot, maybe you wanna take a look a this  [short video](https://www.youtube.com/watch?v=R8rmfD9Y5-c&ab_channel=WebDevSimplified). Focus on the methods we've seen. 
 And, please, if you have any other question... you know where to find us😉

Keep it up💪
Alessandra

PR #2212, opened 2023-06-20
[FTRMT 052023] gokceBerndt

Gokceber:
done


TA-Remote:
hey @Gokceber 
very well done🤩

little detail you could add in your 'randomPick' fuction: 
the remaining contacts are changing every time you are picking one, so you need to update the remainin too, you can go in different way, here some suggestions:
- store the remaining in a state, inside the function use splice, set the new value using setState
- use filter on all contacts removing what is displayed at that moment

keep it up💪
Alessandra

PR #2213, opened 2023-06-20
[FTRMT052023] Fernando

fndoRamos:


TA-Remote:
hey @fndoRamos 
greate and clean solution🤩

love that you make disappear the button to don't get an error!!
another way could also be to have an if statement (checking for the same thing) but inside your function.

keep it up💪
Alessandra

PR #2242, opened 2023-09-02
[PT] María García

MariaG6:


src/App.js
```diff
@@ -1,23 +1,122 @@
...
+        <button className="btn primary" onClick={sortByName}>
+          Sort by name
+        </button>
+        <button className="btn primary" onClick={sortByPopularity}>
+          Sort by popularity
+        </button>
+      </div>
+      {contacts.map((oneContact) => {
+        return (
+          <div>
```
iPoolito, line 68: Remember when we use .map its oportant to add the key prop so react can difference which node you are refering
```suggestion
          <div key={oneContact.id}>
```

iPoolito:
Nice job, keep it like that 🚀

PR #2253, opened 2023-09-09
done

Enzo-Mazzei:


src/App.js
```diff
@@ -1,23 +1,97 @@
...
+                    🏆
+                  </span>
+                ) : null}
+              </td>
+              <td>
+                <button onClick={() => removeContact(contact.id)}>
+                  Delete
+                </button>
+              </td>
+            </tr>
```
iPoolito, line 91: This hole section could be a new component that only handles the information of the contact, in this case you could do something like ` <ContanctInfo  contact={contact} />`

PR #2254, opened 2023-09-09
[PT] Franklin

franklinosei:


src/App.js
```diff
@@ -1,23 +1,93 @@
...
+  const contactsList = contactsData.slice(0, 5);
+
+  const [contacts, setContacts] = useState(contactsList);
+  const [remainingContacts, setRemainingContacts] = useState(contactsData.slice(5));
+
+
+  const addRandomContact = () => {
+    if (remainingContacts.length === 0) {
+      alert("No more contacts to ad`d!");
+      return;
```
iPoolito, line 15: Great job to prevent posible bugs 🔥

src/App.js
```diff
@@ -1,23 +1,93 @@
...
+                <td>{contact.name}</td>
+                <td>{contact.popularity.toFixed(2)}</td>
+                <td>{contact.wonOscar ? <span>🏆</span> : null}</td>
+                <td>{contact.wonEmmy ? <span> 🌟</span> : null}</td>
+                <td>
+                  <button onClick={() => deleteContact(contact.id)}>
+                    Delete
+                  </button>
+                </td>
+              </tr>
```
iPoolito, line 86: This hole section could be a new component that only handles the information of the contact, in this case you could do something like  
```suggestion
          <ContanctInfo  contact={contact} />
```

PR #2256, opened 2023-09-11
Xavier Ortega

Xavieros03:


src/App.js
```diff
@@ -1,24 +1,70 @@
-import logo from './logo.svg';
+
+import { useState } from 'react';
 import './App.css';
+import originalContacts from "./contacts.json"
 
 function App() {
+  const fiveContacts = originalContacts.slice(0, 10)
```
iPoolito, line 7: We need to be consistent with the name of the variables in this cade should be 5
```suggestion
  const fiveContacts = originalContacts.slice(0, 5)
```

src/App.js
```diff
@@ -1,24 +1,70 @@
...
+
+          </tr>
+        </thead>
+        <tbody>
+          {contacts.map(contact => (
+            <tr key={contact.id}>
+              <td><img src={contact.pictureUrl} alt={contact.name} /></td>
+              <td>{contact.name}</td>
+              <td>{contact.popularity.toPrecision(4)}</td>
+              <td>{contact.wonOscar && "🏆"}</td>
```
iPoolito, line 60: In order to give more infromation we can do something like this , so it wont be an empty value
```suggestion
          <td> {contact.wonOscar ? '🏆' : '❌​'}</td>
```

PR #2289, opened 2023-11-01
[WD_OCT23_FT] Guilherme Silva

51V3:


webLisbonTA:
Hi Guilherme, 
Great job on your lab!

Just note that on the Sort by name and popularity you are picking the entire array and the goal was to be pick from the array of 5 that we have initially created.

Overall, good job!

PR #2297, opened 2023-11-01
[LIS-FT-OCT23] Bruno Mendes

brrrun:


webLisbonTA:
Hi Bruno, 
Good job on the lab.

On the random contact it would be important for you to create a new variable for the ramaining contacts:

` const [remainingContacts, setRemainingContacts] = useState(
    contactsData.slice(5, contactsData.length)
  );`

With this code, you would have the remaining contacts with the index 5 included, until the end of the lenght of the contacts.
Then, you would have to create the function random Contacts:


```
function addRandomContact() {
    // First, make a copy of the remaining contacts to prevent changing the original array
    const updateRemaining = [...remainingContacts];

//Then, pick a random index number:
    let randomNum = Math.floor(Math.random() * remainingContacts.length);
    // remove a random element from the copy
    let randomContact = updateRemaining.slice(randomNum, randomNum + 1)[0];

//On the above variable, and as you have only 1 random contact, you will do a slice, from this contact, until this //contact(remember it is only 1) and access the object through [0] - as it is only one contact, it is the 0 Index contact.
//If this is confusing, do a console.log do see the randomContact variable.

    // Note: splice is used to remove an element from the updateRemaining array based
    // on the randomNum index. This element is stored in randomContact.
    // The splice method here removes one element starting at the randomNum index
    // returns an array with that removed element.
    // Accessing the [0] index retrieves the actual removed element.

    // add that contact to a new array
    const updatedContacts = [...contacts, randomContact];
    // update the contacts and the remaining contacts with the corresponding arrays
    setContacts(updatedContacts);
    setRemainingContacts(updateRemaining);
  }
```

PR #2298, opened 2023-11-02
[FT-WD-OCT-2023] João Pedro Silva

joaosilvapedro357:


jacobviv (commented):
Hello João,
don't know if yesterday was a bad day or what happened. Surely today's lab will be better!

src/App.jsx
```diff
@@ -0,0 +1,92 @@
...
+          <tr>
+            <th>picture</th>
+            <th>name</th>
+            <th>popularity</th>
+            <th>Won an Oscar</th>
+            <th>Won an Emmy</th>
+          </tr>
+        </thead>
+        <tbody>
+          {firstContacts.map((contact, index)=>(
```
jacobviv, line 38: this `firstContacts` doesn't exist, I have to change it to contacts to make it work, but then when adding a new contact it is always from the first 5 array

PR #2306, opened 2023-11-08
[FTRMT2023] Marat Pomeranets

pomerama:
Strange problem: if I use the count, the re-rendering of changes done with setContactsArray works ok.
If I dont use count, the contactsArray is being updated correctly (checked e.g. with console.log), but doesnt get re-rendered.


TA-Remote:
hello @pomerama 🙂

The behaviour is due to the way you’re creating a copy of the state variable ‘contactsArray’. 
`
let tempArray = contactsArray;
`
Obj, and array are mutable, and then store in memory by reference, not by value. To copy their data we need to use other methods. Too many to remember them all: https://www.freecodecamp.org/news/how-to-clone-an-array-in-javascript-1d3183468f6a/

I personally like the spread operator, as it can also be used to directly add elements to the copy (at the beginning or at the end)  const copy = [newObj, …stateVariable] —> instead of create copy then copy.unshift(newObj) or .push(). I’m sure the other methods have a better use for other purposes🤔..

```
const tempArray = [...contactsArray];
tempArray.push(randomContact);
// const tempArray = [ …contactsArray, randomContact ];
setContactsArray(tempArray);
```

Here a suggestion to follow conventions😉: 

https://stackoverflow.com/questions/46553062/what-is-the-preferred-declaration-convention-for-objects-or-arrays-const-or-let

I also wanna thank you for your questions, they make me go deeper in concepts/conventions/“code rules” that sometimes I’m just accepting🙃

pomerama:
Hi and many thanks! @TA-Remote  :)
it totally makes sense, and I should know better.. but this is how learning works I guess, hopefully I wont make the same mistake again:) hehe, glad my questions have a positive effect as well:)

PR #2350, opened 2024-01-20
lab-react-ironcontacts- Irene & Jorge

PurrfectNene:


src/App.jsx
```diff
@@ -0,0 +1,94 @@
...
+
+  function randomContacts(){
+    let copiedArray = [...contacts]
+    let indiceAleatorio = Math.floor(Math.random() * contactsData.length)
+    let randomElement = contactsData[indiceAleatorio]
+    if (!copiedArray.some(contact => contact.id === randomElement.id)) {
+      copiedArray.push(randomElement);
+      setContacts(copiedArray);
+    } else {
+      randomContacts();
```
iPoolito, line 18: interesting way to do it 🤯

src/App.jsx
```diff
@@ -0,0 +1,94 @@
...
+import contactsData from"./contacts.json"
+import { useState } from "react";
+
+
+function App() {
+  const [contacts,setContacts] = useState(contactsData.slice(0,5))
+  console.log(contacts)
+
+  function randomContacts(){
+    let copiedArray = [...contacts]
```
iPoolito, line 11: We need a guard safe in order to not add more contacts that we have could be something like this 

```suggestion
   if (contactsData.length === contacts.length) {
      console.log("No more contacts can be add")
      return
    }
    let copiedArray = [...contacts]
```

iPoolito:
Nice jooob 🚀  🔥

PR #2352, opened 2024-01-21
Eleanor & Christian IronContacts Lab

eleanorpitcher:


src/App.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+
+function App() {
+
+  const contactList = Contacts.slice(0,5)
+  const remainingContacts = Contacts.slice(5)
+  // console.log(remainingContacts)
+  // console.log(contactList)
+
+  const [contacts, setContacts] = useState(contactList)
```
iPoolito, line 14: Here in this case in order to no get repeated contacts we can do something like this 

```suggestion
  const contactList = Contacts.slice(0, 5)

  const [contacts, setContacts] = useState(contactList)
  const remainingContacts = Contacts.filter((mainDataContacts) => !contacts.some((currentContactFromList) => mainDataContacts.id == currentContactFromList.id))
```
What we are doing here is filtering from the Contacts array that is the JSON file, the items that doenst have the same id form our contacts state

src/App.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+  function getRandomContact(){
+    const copyContacts = [...contacts]
```
iPoolito, line 17: Here we need just a guard in order to stop adding when the remaing are 0

```suggestion
  if (remainingContacts.length === 0) {
      console.log("No more contacts can be add")
      return
    }
    const copyContacts = [...contacts]
```

iPoolito:
Good joob congratz 🔥 🚀

PR #2354, opened 2024-01-21
[WD-PT] Juan Morales and Mónica Agrela - IronContacts

Juanmormar:


src/App.js
```diff
@@ -1,25 +1,83 @@
...
-    </div>
-  );
+
+const initialcontacts = contactsData.slice(0,5)
+const [contacts, setContacts] = useState(initialcontacts)
+
+// funtions
+
+function addRandomContact(){
+const contactsCopy = [...contacts]
```
iPoolito, line 14: Niceee joob only we need a safe guard when we no have more contacts to add 

```suggestion
function addRandomContact(){
   if (contacts.length === contactsData.length) {
      console.log("No more contacts to add")
      return
    }
const contactsCopy = [...contacts]
```

Juanmormar:
Great to know!

Gracias Pool!
________________________________

PR #2356, opened 2024-01-23
[WD-PT] Diogo Moreira & Nadine Rangosch

nenuphars:


src/App.jsx
```diff
@@ -0,0 +1,85 @@
...
+      });
+  
+      if (!isDuplicate) {
+        contactsCopy.push(contactsData[randomNum]);
+        setContacts(contactsCopy);
+        break;
+      }
+    } while (true);
+  
+  }
```
iPoolito, line 24: The thing here is that we are waiting that the random number is not duplicated so probably it will take us a lot of time in order to wait to get the last contacts that are not duplicated so what we can do is  create another variable call remainingContacts  that will help us whit the filter something like this 

```suggestion
//First check if the remainingContacts array length its not 0  so we cant add morie contacts
       if (remainingContacts.length === 0) {
      console.log("More contacts cant be added")
      return
    }
    let contactsCopy = [...contacts]
//here we gets the random index from the remainingContacts array so we always are going to get at ID that corresponds to the ones thar are not in the current List
    let randomNum = Math.floor(Math.random() * remainingContacts.length)
//Get the new contact with the random index
    const newContact = remainingContacts[randomNum]
    //Push the new contact to the copy array
    contactsCopy.push(newContact)
    //remvoe the new contact for the remaingArray because now its in the list
    remainingContacts.slice(randomNum, 1)
    //updates the state
    setContacts(contactsCopy)
  }
```

src/App.jsx
```diff
@@ -0,0 +1,85 @@
+import contactsData from './contacts.json'
+import { useState} from 'react'
+import "./App.css";
+
+function App() {
+  const [contacts, setContacts] = useState(contactsData.slice(0,5))
+
```
iPoolito, line 7: Here is the initial state of remainingContacts 
```suggestion
  const remainingContacts = contactsData.filter((comingContact) => !contacts.some((contact) => comingContact.id == contact.id))
```

